PR #5, opened 2023-06-14
fix: Marshal yaml with 2 space indents

alexstojda:
- Also return error when no releases found

### DIFF
--- a/internal/utils/yaml.go
+++ b/internal/utils/yaml.go
@@ -1,8 +1,11 @@
 package utils
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"gopkg.in/yaml.v3"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


internal/utils/yaml.go
```diff
@@ -1,8 +1,11 @@
 package utils
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"gopkg.in/yaml.v3"
```
riclima, line 7: FIY, but I discovered yesterday while playing around with making YAML patches using Go that yaml.v3 [does not format lists like yaml.v2 ](https://github.com/kubernetes-sigs/kustomize/issues/3946) (i.e., Kubernetes), with no option to go back to the old setting. Reasoning from the devs is that the old way had weird edge cases and new way doesn't, so "breaking" change in v3.

Old:

```yaml
favorites:
  fruit: apple
  car: ford
fruits
- apple
- orange
```

New:

```yaml
favorites:
  fruit: apple
  car: ford
fruits
  - apple
  - orange
```

I like new better. Makes more sense, but just a random FIY.

riclima (approved):

internal/utils/yaml.go
alexstojda: Yeah I like the new formatting better too, minus their decision to default the indents to 4 spaces lol.

PR #13, opened 2023-07-25
Add sealed-secrets integration

silphid:
- Add `joy secret import` to fetch sealed-secrets certificate from cluster and store it into Environment CRD.
- Add `joy secret seal` to encrypt a secret using selected Environment CRD.

### DIFF
--- a/cmd/joy/secret.go
+++ b/cmd/joy/secret.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"github.com/nestoca/joy/internal/secret"
+	"github.com/spf13/cobra"
+)
+
+func NewSecretCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "secret",
+		Aliases: []string{"secrets", "sec"},
+		Short:   "Manage sealed secrets",
+		Long: `Manage sealed secrets, such as sealing (encrypting) secrets and importing public certificate from cluster.
+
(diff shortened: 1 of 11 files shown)


riclima (approved):

cmd/joy/secret.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/nestoca/joy/internal/secret"
+	"github.com/spf13/cobra"
+)
+
+func NewSecretCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "secret",
+		Aliases: []string{"secrets", "sec"},
```
alexstojda, line 11: I wonder if we could have a better name for this that makes it more obvious we're managing sealed-secrets and not just arbitrary secrets. Yes it says it in the description, but who reads the manual these days xD 

Maybe `sealed-secrets` with aliases `ss` and `ssecrets` ?

alexstojda (commented):
no tests? 👀

cmd/joy/secret.go
silphid: Yeah I'm on the fence for this one, I get your point and I partly agree.  I still love simple and straight to the point command names, so I would propose we make `sealed-secrets` the official name of the command, that people will see when they do `joy --help`, however still provide `secret`, `secrets` and `sec` aliases.

PR #19, opened 2023-08-31
Fix mixed up source and target environments in `joy release promote`

silphid:
Some past refactoring had introduced that bug, where `joy release promote` would load environments in any arbitrary order in which they were read from disk, instead of respecting the expectation that first environment must be the source and second the target, which is assumed by the rest of promotion logic.


alexstojda (changes requested):
Could you add a test to check for the regression of this bug? :) 

LGTM otherwise!

alexstojda (approved):

PR #22, opened 2023-09-11
feat(devops-1262): add `joy pr promote` command

silphid:
https://nestoca.atlassian.net/browse/DEVOPS-1262


internal/pr/promote/promotion.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("getting promotion environment for branch %s pull request: %w", branch, err)
+	}
+
+	promotableEnvironmentNames := getPromotableEnvironmentNames(environments)
+	env, err = p.Prompt.WhichEnvironmentToPromoteTo(promotableEnvironmentNames, env)
+	if err != nil {
+		return fmt.Errorf("prompting user to select promotion environment: %w", err)
+	}
+
+	if err := p.PullRequestProvider.SetPromotionEnvironment(branch, env); err != nil {
```
j-martin, line 80: Assuming there is a PR with the same `promotion` label, should we have a check to confirm it is ok to steal the label?

internal/pr/promote/github_pull_request_provider.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return fmt.Errorf("getting promotion labels for branch %s: %w", branch, err)
+	}
+
+	// Remove existing labels, if any
+	for _, label := range labels {
+		if err := removeLabel(branch, label); err != nil {
+			return fmt.Errorf("removing label %s from branch %s: %w", label, branch, err)
+		}
+	}
```
j-martin, line 103: Should we also delete the promotion label on whichever PR may have the same promotion environment?

It could be confusing if two PRs have the same target. Which one would work?

j-martin (commented):
My PR is mostly questions, otherwise, things are looking good.

cmd/joy/git.go
```diff
@@ -1,18 +1,34 @@
...
-		Short:   "Call arbitrary git command in catalog dir with given arguments",
-		Long:    `Call arbitrary git command in catalog dir with given arguments`,
-		GroupID: "git",
-		Args:    cobra.ArbitraryArgs,
+		Use:                "git",
+		Short:              "Call arbitrary git command in catalog dir with given arguments",
+		Long:               `Call arbitrary git command in catalog dir with given arguments`,
+		GroupID:            "git",
+		Args:               cobra.ArbitraryArgs,
+		DisableFlagParsing: true,
```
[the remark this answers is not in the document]
silphid: I know, I know! 😁

internal/pr/promote/github_pull_request_provider.go
```diff
@@ -0,0 +1,131 @@
...
+		label := fmt.Sprintf("promote:%s", env)
+		if err := addLabel(branch, label); err != nil {
+			return fmt.Errorf("adding label %s to branch %s: %w", env, branch, err)
+		}
+	}
+	return nil
+}
+
+func removeLabel(branch string, label string) error {
+	cmd := exec.Command("gh", "pr", "edit", branch, "--remove-label", label)
```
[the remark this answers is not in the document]
silphid: Yeah, definitely, I'm already doing a check for other dependencies, such as `jac` and `kubeseal` (Sealed Secrets), I will add same here.

cmd/joy/pr.go
```diff
@@ -0,0 +1,43 @@
...
+	cmd.AddCommand(NewPRPromoteCmd())
+	return cmd
+}
+
+func NewPRPromoteCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:     "promote",
+		Aliases: []string{"prom"},
+		Short:   "Auto-promote builds of pull request to given environment",
+		Args:    cobra.NoArgs,
```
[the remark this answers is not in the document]
silphid: Definitely agreed. It's just a design decision we made to first deliver a MVP that is fully interactive, thus as easy as possible to learn and use, and then add 100% support for everything via command line arguments/flags. It's just a matter of time! 🙂

go.mod
```diff
@@ -5,12 +5,15 @@ go 1.19
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/TwiN/go-color v1.4.0
+	github.com/go-test/deep v1.1.0
+	github.com/golang/mock v1.6.0
```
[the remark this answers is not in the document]
silphid: Good to know! It's ChatGPT that directed me to it, but then ChatGPT's model is itself soooo deprecated time-wise, except for everything related to software! 😜

internal/pr/promote/survey_prompt.go
```diff
@@ -0,0 +1,61 @@
+package promote
+
+import (
+	"fmt"
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/nestoca/joy/internal/style"
+)
+
+type SurveyPrompt struct {
```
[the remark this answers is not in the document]
silphid: Yeah, you're right, especially as you said that it also handles outputs, not just survey prompts.  However, I would probably call the implementation `InteractivePromptProvider`.

internal/pr/promote/promotion.go
silphid: Hum, very interesting! I think that would be feasible, especially as the exclusivity scope would typically be just that same repo. So we could list all PRs and check labels. If we find another one, we could prompt the user for confirmation, and upon positive response delete the other label.

internal/pr/promote/github_pull_request_provider.go
silphid: Indeed, they would be fighting against each others at every one of their builds.  Per your above comment, let's try to implement that!

go.mod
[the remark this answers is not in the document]
silphid: I just migrated to the uber version.

internal/pr/promote/github_pull_request_provider.go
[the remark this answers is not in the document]
silphid: Added dependency check

internal/pr/promote/mock_pull_request_provider.go
```diff
@@ -0,0 +1,92 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
silphid: Removed mocks, git ignored them, added installation of mockgen to makefile (and thus indirectly also to CI workflow)

internal/gh/gh.go
```diff
@@ -9,13 +10,28 @@ import (
...
 }
+
+func ensureGHInstalled() error {
+	cmd := exec.Command("which", "gh")
+	err := cmd.Run()
+	if err != nil {
+		fmt.Println("🤓 This command requires the gh cli.\nSee: https://github.com/cli/cli")
+		return errors.New("missing gh cli dependency")
+	}
+	return nil
```
j-martin, line 36: I was thinking about something like this untested code:

```go
func ensureGHInstalled() error {
	return ensureToolIsInstalled("gh", "This command requires the gh cli.\\nSee: https://github.com/cli/cli")
}

func ensureToolIsInstalled(toolName string, extraHelp string) error {
	cmd := exec.Command("which", toolName)
	err := cmd.Run()
	if err == nil {
		return nil
	}
	if runtime.GOOS == `darwin` {
		fmt.Printf("Installing '%s' with brew\n", toolName)
		cmd = exec.Command("brew", "install", toolName)
		err = cmd.Run()
	}
	if err != nil {
		fmt.Printf("Ensure '%s' is in your PATH. %s\n", toolName, extraHelp)
		return errors.New(fmt.Sprintf("missing %s dependency", toolName))
	}
	return nil
}

```

silphid: Very interesting!  However, not sure about automatic "lazy" installation when the command is run...  Would maybe more implement that in `joy setup`, especially as some dependencies like `jac` require further setup (clone people repo).  Maybe we could suggest user to run `joy setup` instead, or prompt whether to perform that installation (and then execute just that portion of the setup process)? I will create a separate ticket for that, instead of adding all that logic within this PR (following advice from a great collaborator who just joined our team!)

silphid: @j-martin please review that ticket and let me know what you think:
https://nestoca.atlassian.net/browse/DEVOPS-1938

silphid: Here's the ticket! 🙂
https://nestoca.atlassian.net/browse/DEVOPS-1938

j-martin: The ticket looks good. I could imagine installing the tools dynamically could be done in this PR, but it is also reasonable not to pack too much in a PR.

That got me thinking about `gh` precisely, now `joy` would also depend on `gh` being configured properly.

We could check the auth status:

```
$ gh auth status
github.com
  ✓ Logged in to github.com as j-martin (keyring)
  ✓ Git operations for github.com configured to use ssh protocol.
  ✓ Token: gho_************************************
  ✓ Token scopes: admin:public_key, gist, read:org, repo
```

silphid: That's what I ended up doing, let me know if that's what you had in mind:
https://github.com/nestoca/joy/commit/ae2dd774b09d058b889811da90945af345888676

j-martin: We really need to check the exit code.

```
$ gh auth status
github.com
  ✓ Logged in to github.com as j-martin (keyring)
  ✓ Git operations for github.com configured to use ssh protocol.
  ✓ Token: gho_************************************
  ✓ Token scopes: admin:public_key, gist, read:org, repo
$ echo $?
0

# Deleted the token from my local keychain
$ gh auth status
github.com
  X github.com: authentication failed
  - The github.com token in /Users/jm/.config/gh/hosts.yml is no longer valid.
  - To re-authenticate, run: gh auth login -h github.com
  - To forget about this host, run: gh auth logout -h github.com
$ echo $?
1
```

silphid: The code actually works, because it checks for the presence of `Token: XXXXXX`.  I ignored exit code / error, to focus on that check instead, because the two checks were redundant, but I could rely solely on `gh auth status` returning 0 exit code (plus the scopes check).

internal/pr/promote/github_pull_request_provider.go
```diff
@@ -112,6 +110,44 @@ func (g *GitHubPullRequestProvider) SetPromotionEnvironment(branch string, env s
...
+	// We can safely assume that there is either none or only one PR for a given branch
+	if len(prs) == 0 {
+		return nil, nil
+	}
+	return &prs[0], nil
+}
+
+func getAll() ([]pullRequest, error) {
+	// List pull requests for branch
+	cmd := exec.Command("gh", "pr", "list", "--json", "headRefName,labels")
```
j-martin, line 141: On older/more active repos, we are going to hit paging issues. We can filter out by labels:

```
$ gh pr list --label 'promote:<env>' --state open
```

We might want to get merged or closed PRs but remove the labels without prompting the user.

internal/pr/promote/github_pull_request_provider.go
```diff
@@ -14,30 +14,10 @@ type GitHubPullRequestProvider struct {
 var labelRegex = regexp.MustCompile(`^promote:\s*(\w+)$`)
 
 type pullRequest struct {
-	Labels []struct {
+	HeadRefName string `json:"headRefName"`
+	Labels      []struct {
```
j-martin, line 19: I am wondering if we should use the proper github go library even if we just stole the `gh` auth token from the keychain? There is already a fair amount of dancing around (parsing stdout) of the output of `gh`.

image


https://github.com/cli/cli/blob/trunk/internal/keyring/keyring.go#L34C7-L53

internal/pr/promote/interactive_prompt_provider.go
```diff
@@ -0,0 +1,82 @@
...
+		Default: preSelectedEnv,
+	}
+	var selectedEnv string
+	err := survey.AskOne(prompt, &selectedEnv)
+	if err != nil {
+		return "", fmt.Errorf("prompting user to select environment: %w", err)
+	}
+	if selectedEnv == none {
+		selectedEnv = ""
+	}
```
j-martin, line 43: In which case we would want `"[none]"` / `""`?

internal/gh/gh.go
```diff
@@ -26,12 +28,31 @@ func CreatePullRequest(args ...string) error {
...
 		fmt.Println("🤓 This command requires the gh cli.\nSee: https://github.com/cli/cli")
 		return errors.New("missing gh cli dependency")
 	}
+
+	cmd = exec.Command("gh", "auth", "status")
+	output, _ := cmd.CombinedOutput()
+	outputStr := string(output)
+
+	// Check if authorized
+	tokenRegex := regexp.MustCompile(`Token: .+`)
```
riclima: FIY, when you run `gh auth status` it returns an error code 1 if you're not authenticated. You don't need to regex the output, you can do like line 34 and check for errors. If not nil, then you're not authenticated.

internal/gh/gh.go
```diff
@@ -26,12 +28,31 @@ func CreatePullRequest(args ...string) error {
...
+	if !tokenRegex.MatchString(outputStr) {
+		fmt.Printf("🔐 Please run %s to authorize the gh cli.\n", style.Code("gh auth login"))
```
riclima, line 44: `auth` in the context of the `gh` tool is actually authenticate (verifying who you are). Authorization is checked when you try to do operations (e.g.: trying to clone a repo that you don't have permission to access).

riclima (approved):

internal/gh/gh.go
```diff
@@ -26,12 +28,31 @@ func CreatePullRequest(args ...string) error {
 	return nil
 }
 
-func ensureGHInstalled() error {
+func EnsureInstalledAndAuthorized() error {
 	cmd := exec.Command("which", "gh")
```
riclima, line 32: Consider:

```
cmd := exec.Command("command", "-v", "gh")
```

[Apparently](https://stackoverflow.com/a/677212), `command -v` is more reliable as `which` can have weird behaviour, including not setting an exit status. This would cause the following lines to not work correctly.

riclima, line 32: Consider:

```
cmd := exec.Command("command", "-v", "gh")
```

[Apparently](https://stackoverflow.com/a/677212), `command -v` is more reliable as `which` can have weird behaviour, including not setting an exit status. This would cause the following lines to not work correctly.

internal/pr/promote/github_pull_request_provider.go
silphid: I addressed that, thanks! :)

internal/pr/promote/github_pull_request_provider.go
silphid: That would be great, I initially was hoping to do that, but couldn't figure out how to leverage the user's current authorization via API. However, please create another ticket for that one, as we really need to conclude this one! 😁

internal/pr/promote/interactive_prompt_provider.go
```diff
@@ -0,0 +1,82 @@
...
+	prompt := &survey.Confirm{
+		Message: fmt.Sprintf("Another pull request for branch %s is already auto-promoting to %s environment, disable it?", style.Resource(branch), style.Resource(env)),
+		Default: false,
+	}
+	var shouldDisable bool
+	err := survey.AskOne(prompt, &shouldDisable)
+	if err != nil {
+		return false, fmt.Errorf("prompting user to confirm disabling promotion on other pull request: %w", err)
+	}
+	return shouldDisable, nil
```
j-martin, line 57: In which case would a user say no and still want to continue?

I would assume this prompt would be a confirmation, as in Yes (continue) or No (abort).

j-martin (approved):
I have one question. Otherwise LGTM! :)

PR #27, opened 2023-10-06
feat: lock, handle both # lock and ## lock

j-martin:
I think it is more confusing for that `# lock` is not handled.

### DIFF
--- a/internal/yml/locking.go
+++ b/internal/yml/locking.go
@@ -1,11 +1,12 @@
 package yml
 
 import (
-	"gopkg.in/yaml.v3"
 	"regexp"
+
+	"gopkg.in/yaml.v3"
 )
 
-var lockMarkerRegex = regexp.MustCompile(`(?m)^##\s*lock\s*$`)
+var lockMarkerRegex = regexp.MustCompile(`(?m)^#+\s*lock\s*$`)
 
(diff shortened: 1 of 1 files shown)


internal/yml/locking.go
```diff
@@ -1,11 +1,12 @@
...
 
 import (
-	"gopkg.in/yaml.v3"
 	"regexp"
+
+	"gopkg.in/yaml.v3"
 )
 
-var lockMarkerRegex = regexp.MustCompile(`(?m)^##\s*lock\s*$`)
+var lockMarkerRegex = regexp.MustCompile(`(?m)^#+\s*lock\s*$`)
```
silphid, line 9: Would rather not support an arbitrary number of hash signs! 😁
```suggestion
var lockMarkerRegex = regexp.MustCompile(`(?m)^##?\s*lock\s*$`)
```
And I would assume that this is just temporary, until we implement the tag and deprecate the comment, because just adding more and more ways to do things will only get people even more confused!

silphid (changes requested):

internal/yml/locking.go
j-martin: Yes, while we support the comment. I think `### lock` should also work.

silphid (approved):

PR #28, opened 2023-10-12
fix: preserve order of locked elements

silphid:
Locked elements only present in target yaml were moved to the end of their siblings. We should instead try to preserve their order by looking at keys coming before them and trying to find equivalent insertion position.

### DIFF
--- a/internal/yml/merge.go
+++ b/internal/yml/merge.go
@@ -86,7 +86,14 @@ func mergeSubTrees(src, dest *yaml.Node) *yaml.Node {
 				src.Content[srcIndex+1] = subtree
 			} else {
 				// We are adding a new node
-				src.Content = append(src.Content, destKeyNode, subtree)
+				insertIndex := findInsertionIndex(src, dest, key)
+				if insertIndex != -1 {
+					// Insert at correct position
+					src.Content = append(src.Content[:insertIndex], append([]*yaml.Node{destKeyNode, subtree}, src.Content[insertIndex:]...)...)
+				} else {
+					// Append at the end
+					src.Content = append(src.Content, destKeyNode, subtree)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/yml/merge.go
```diff
@@ -98,6 +105,33 @@ func mergeSubTrees(src, dest *yaml.Node) *yaml.Node {
 	return nil
 }
 
+func findInsertionIndex(src, dest *yaml.Node, key string) int {
```
j-martin, line 108: Not that it matters, but that can't be fast...

internal/yml/merge_test.go
```diff
@@ -380,3 +380,34 @@ e:
...
+  l: m
+`
+	expectedContent := `
+a: b
+c: d
+e:
+  f: g
+  ## lock
+  j: k
+  l: m
```
j-martin, line 409: Out of curiosity, what was the output before the fix? More or less random?

j-martin (approved):

internal/yml/merge.go
silphid: I'm sure you've heard what Confucius had to say about that:
> Premature optimization is the root of all Evil.
That's a `O(n^2)` algorithm, but even a `O(n^3)` or worse can be blazingly fast when `n` is rather small (which will always be the case for Joy releases, even super hefty ones, probably staying in the range of 5 to 50 siblings, tops). Not even considering that merges are performed on user's machine, only a few at a time, with wait time barely perceptible.  I'd say we wait until it becomes an issue (if ever) and then revisit, but I would be very curious and interested to know if you have any alternative algorithm achieving same result more efficiently (even if we don't necessarily implement it right now, just for my bettering! 😁)

internal/yml/merge_test.go
silphid: No, it was predictable, always appended at the end. It's just that it would result in Joy wanting to update such nodes just to move them to the end, resulting in unneeded changes.

internal/yml/merge.go
j-martin: Yep, hence the "Not that it matters" 🙂

PR #35, opened 2023-10-30
chore(formatting): add fmt task to makefile: run gofumpt and goimports

davidmdm:
I noticed that the imports were unorganized (thirdparty + nesto + standard-lib) were all intertwined. 
At previous companies we generally leaned on tools such as gofumpt (compliant gofmt formatter but stricter) and goimports to format and organize our imports.

Changes:
- added `fmt` task to makefile
  - run `gofumpt`
  - run `goimports`
- ran fmt task

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,3 +27,7 @@ test-cov: generate vet
 
 clean:
 	@rm -rf ./reports ./out
+
+fmt:
+	@gofumpt -w .
+	@goimports --local github.com/nestoca/joy,github.com/nestoca -w .
--- a/api/v1alpha1/environment.go
+++ b/api/v1alpha1/environment.go
@@ -2,10 +2,12 @@ package v1alpha1
 
 import (
 	"fmt"
-	"github.com/nestoca/joy/internal/yml"
+	"path/filepath"
+
 	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v3"
-	"path/filepath"
+
+	"github.com/nestoca/joy/internal/yml"
 )
 
 const EnvironmentKind = "Environment"
--- a/api/v1alpha1/release.go
+++ b/api/v1alpha1/release.go
@@ -2,8 +2,10 @@ package v1alpha1
 
 import (
 	"fmt"
-	"github.com/nestoca/joy/internal/yml"
+
 	"gopkg.in/yaml.v3"
+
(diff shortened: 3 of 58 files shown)


silphid (approved):
@davidmdm beyond applying that formatting via `make`, can we make GoLand also do it automatically (or via a plugin)?

Makefile
```diff
@@ -27,3 +27,7 @@ test-cov: generate vet
 
 clean:
 	@rm -rf ./reports ./out
+
+fmt:
+	@gofumpt -w .
+	@goimports --local github.com/nestoca/joy,github.com/nestoca -w .
```
j-martin, line 33: Would this be equivalent?

```suggestion
	@goimports --local github.com/nestoca -w .
```

Makefile
```diff
@@ -27,3 +27,7 @@ test-cov: generate vet
 
 clean:
 	@rm -rf ./reports ./out
+
+fmt:
+	@gofumpt -w .
```
j-martin, line 32: Can we add a development section to the readme that specifies which versions of `gofumpt` and `goimports` needs to be installed? Alternatively, we could have a `make setup` target that installs those dependencies.

j-martin (commented):
Overall, that's very good change. It is also in line with other team's practices. 🙂

A few minor changes I would make:

- Can we split the commit for the actual change (Makefile, docs, ci, etc.) from the reformatting change? That way, that commit can be quickly ignored later on, and it makes it more apparent what is a human change (right now, just the Makefile) as opposed to an automated change.
- I think we should enforce the formatting in CI, or else we'll end up with a fair amount of formatting trashing between PRs. If you are not familiar with GitHub Actions that's a good starting task. :)

Makefile
davidmdm: Not exactly. Basically the former separates third party imports, joy imports, and nestoca imports.
The suggestion separates third party imports from nesto imports. Both are fine. One is coarser and one is more fine-grained.

j-martin: I see, so they would get grouped in different groups (`nestoca/joy/*` vs `nestoca/*`)? If so, we currently don't make that distinction in our other projects (see [screenshot](https://nestoca.atlassian.net/wiki/spaces/NP/pages/1896906768/Go+Development+Environment#GoLand)), so we may want to keep it the same (my suggestion above).

PR #36, opened 2023-10-31
fix(gh): error should output combined stdout and stderr

silphid:


davidmdm (approved):

internal/git/pr/github/gh.go
```diff
@@ -29,12 +29,9 @@ func executeInteractively(args ...string) error {
 	}
 
 	cmd := exec.Command("gh", args...)
-	cmd.Stdin = os.Stdin
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	err = cmd.Run()
+	output, err := cmd.CombinedOutput()
```
j-martin, line 32: The comment on L24 is not true anymore.

j-martin (approved):
I don't have strong feelings either way, but this change either hides  `gh` output in the happy path or shows the output at the end of a failure.

I would personally privilege surfacing `gh` output as it happens (as it was before this PR) to give feedback to the user that at least something is happening. As opposed to have `joy` seemingly "hangs" while `gh` is doing its thing.

That being said I am fine either way.

PR #40, opened 2023-11-03
fix: pin build dependencies

j-martin:
Or else there is no guarantee what will be executed across different environments.

Also uses the same import convention as other projects

As discussed in https://github.com/nestoca/joy/pull/35#discussion_r1377915933

This is a no-op since joy doesn't use our other packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,4 +32,4 @@ clean:
 
 fmt:
 	@gofumpt -w .
-	@goimports --local github.com/nestoca/joy,github.com/nestoca -w .
+	@goimports --local github.com/nestoca/joy -w .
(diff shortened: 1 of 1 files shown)


j-martin:
Side note, @davidmdm is there a reason why https://github.com/nestoca/joy/pull/35 got closed in favor of https://github.com/nestoca/joy/pull/38?

riclima (approved):

davidmdm:
@j-martin , I closed the original PR, and created two pull requests:
- the tooling change within the makefile
- applying the automated changes

I made the 2nd PR point to the first, however it was approved and merged into the first, and then the first was merged into master. Ideally it should have been done the other way.

j-martin:
I see. In this case, I don't think we needed two PRs. Separating changes by commit is totally fine. 🙂

PR #47, opened 2023-11-13
feat(PL-2089): add min version to joyrc and root validations

davidmdm:
Joy needs a concept of configurable minimum version from the catalog dir, so that we can enforce updates overtime.

Changes:
- added `minVersion` to joyrc configuration
- give precedence to catalog for min version configuration over home joyrc
- added validations in root prerun for min version being satisfied
- added prompt when running a build version with a hidden flag for skipping prompt.


riclima (approved):

cmd/joy/root.go
```diff
@@ -0,0 +1,98 @@
...
+
+			if !skipVersionCheck {
+				if version != "" && semver.Compare(version, cfg.MinVersion) < 0 {
+					return fmt.Errorf("current version %q is less than required minimum version %q. Please update joy", version, cfg.MinVersion)
+				}
+				if !semver.IsValid(version) {
+					var ok bool
+					prompt := &survey.Confirm{Message: "you are running joy on a development build. Do you wish to continue?"}
+					if err := survey.AskOne(prompt, &ok); err != nil || !ok {
+						return err
```
davidmdm, line 58: if ok is false need to return a non-nil error

internal/dependencies/dependency.go
```diff
@@ -22,8 +22,7 @@ type Dependency struct {
 }
 
 func (d *Dependency) IsInstalled() bool {
-	cmd := exec.Command("command", "-v", d.Command)
-	return cmd.Run() == nil
+	return exec.Command("/bin/bash", "-c", "command -v "+d.Command).Run() == nil
```
riclima, line 25: Heya, what's the reasoning for this change?

riclima: Unless I'm missing something, `command` is a POSIX required utility so it should be available in just about any UNIX environment, and in this new version we require `/bin/bash` which could not be available in some environments (notably, probably macOS one day, as the version that it ships is extremely old and apparently they won't update due to GPL 3), as well as if we ever use this in some automated environment and `d.Command` for whatever reason becomes possible to be set by a user, we would have a code injection.

riclima: It also opens the potential for [weird side-effects](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html) if `BASH_ENV` is set.

davidmdm: In github actions, the default shell does not have `command`. 

`exec.Command("command", ...)` returns an error: `exec: "command": executable file not found in $PATH`.

So to guard against this we are explicitly starting a bash session.

riclima: Wow that's odd. Can you try this?

```suggestion
	return exec.Command("/usr/bin/env", "bash", "-c", "command", "-v", d.Command).Run() == nil
```

This guards against using really old bash version in macOS or it not being available altogether from the system but being available as a package from elsewhere. I'm not sure if passing `d.Command` as its own arg protects it against it being a code injection though, so I'll check it out later.

riclima: Does this work?

```suggestion
	_, err := exec.LookPath(d.Command)
	return err == nil
```

PR #50, opened 2023-11-14
chore(packages): add deb package generation in goreleaser

damethot:
Went ahead and added debian package generation in goreleaser via [nfpm](https://goreleaser.com/customization/nfpm/).
@silphid I wasn't sure about which email I should put in the maintainer field, or we could opt to omit this property altogether.

Output of a local goreleaser run:
image

Locally installing the resulting deb package
image

Validating proper installation
image

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -8,7 +8,7 @@ builds:
       - linux
       - darwin
     ldflags:
-      - -X "main.version=v{{ .Version }}"
+      - -X "main.version={{ .Version }}"
 archives:
   - format: tar.gz
     wrap_in_directory: false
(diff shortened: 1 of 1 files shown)


silphid (approved):

damethot:
Seems like there's a little issue with the semver, the `joy version` command returns "vvX.X.X". I'll investigate.

damethot:
@silphid There's an extra "v" character in front of the version injection when compiling the `joy` binary, is this intentional?
https://github.com/nestoca/joy/blob/master/.goreleaser.yaml#L11

I've ran locally with it removed and, as far as I can tell, the generated binaries still yield `v0.18.1` from `joy version`. I can't test the darwin target to see if that one still contains it though.

davidmdm (approved):

damethot:
I've asked MDP to run the built darwin binary and it still has the "v" prefix. So I pushed the change to remove the "v" prefix in the version injection.
image

silphid:
@damethot @davidmdm I'm a bit confused here... I've always had a single `v`:
```
$ joy version
v0.16.3
```
Since when has it become doubled? Maybe something we changed in our joy GHA workflow?

PR #62, opened 2023-12-13
fix(PL-2148): tag inference should ignore custom tags

davidmdm:


pkg/catalog/tags.go
```diff
@@ -66,6 +66,7 @@ var knownsTags = []string{
 	"!!map",
 	"!!str",
 	"!!seq",
+	"!!null",
 
 	// Custom tags
 	"!lock",
```
silphid, line 72: Var should probably be called `knownTags` (while we are at it!)

silphid (approved):
Just a minor var rename suggestion, but LGTM! 👍

pkg/catalog/tags.go
davidmdm: typos! They are everywhere :\

silphid (approved):

PR #63, opened 2023-12-14
fix(PL-2156): fix promoted filepath to be the same relative to source

davidmdm:
JIRA: https://nestoca.atlassian.net/browse/PL-2156


alexstojda (approved):
Amazing, thanks!

silphid (approved):

PR #66, opened 2023-12-22
feat(PL-2051): improve joy release promote UX for promotable changes

silphid:


internal/release/cross/release.go
```diff
@@ -21,6 +20,9 @@ type Release struct {
 	// are respectively source and target. If merged result is same as target, then no promotion is needed and
 	// PromotedFile is nil. This must be explicitly computed via ComputePromotedFile().
 	PromotedFile *yml.File
+
```
davidmdm, line 23: Can all the changes in this file be reverted? I don't think these boolean are used anymore? I could be wrong.

internal/release/cross/release.go
```diff
@@ -62,7 +66,14 @@ func (r *Release) ComputePromotedFile(sourceEnv, targetEnv *v1alpha1.Environment
 	}
 
 	// Only consider promotion if the new merged result is different from existing target
-	if targetRelease == nil || !yml.Compare(targetRelease.File.Yaml, promotedFile.Yaml) {
+	if targetRelease != nil {
```
davidmdm, line 69: Oh nevermind... I thought we were going to determine if a revision is in sync at the time we need that information. I have a slight preference for colocating this logic to where we need it and keep ComputePromotedFile as simple as possible (IE only computes promotedFile).

If you prefer it like this that's fine.

internal/release/promote/interactive_prompt_provider.go
```diff
@@ -61,11 +61,18 @@ func (i *InteractivePromptProvider) SelectReleases(list *cross.ReleaseList) (*cr
...
-		choice := fmt.Sprintf("%s\t%s\t>\t%s\t",
-			style.Resource(item.Name),
-			cross.GetReleaseDisplayVersion(item.Releases[1], inSync),
-			cross.GetReleaseDisplayVersion(item.Releases[0], inSync))
+		inSync := item.VersionInSync && item.ValuesInSync
+		var choice string
+		if inSync {
+			choice = fmt.Sprintf("%s\t%s",
+				style.PromotableRelease(item.Name, false),
+				GetReleaseDisplayVersion(item.Releases[1], true, true, true))
```
davidmdm, line 69: Heated topic, but line 69 intention is now unknowable by directly reading it unless you have that function's context in your mind, the files are open side by side, or you use editor capabilities to inspect it.
I would strongly encourage to decouple content from style!

internal/yml/compare.go
```diff
@@ -38,3 +38,51 @@ func deepEqualNode(node1, node2 *yaml.Node) bool {
...
+
+	if a.Kind != b.Kind || a.Value != b.Value {
+		return false
+	}
+	if len(a.Content) != len(b.Content) {
+		return false
+	}
+
+	// Determine current key and rest of path
+	excludedKey := ""
```
davidmdm, line 66: There's a couple issues here!
- I am finding it hard to follow this logic
- I am not able to prove to myself that its correct
- (Not important because its not our current need but...) This approach doesn't support sequences

On line 72 we remove a level from the excluded path, and pass it to line 82 regardless of if the key on the current level matches the removed index of the excluded path.

Therefore I think this actually ignores mapped items have the same amount of nesting as the given path and that have the same key as the last segment of the given path.

Also line 76 isn't compatible with nodes that have the empty string as their key even though that's an edge case we won't see, we can still consider it for correctness. 

Given that it is already hard to follow for my small brain, I wouldn't try and patch it with more if statements... Rather I would suggest you have a private implementation that recursively checks the logic and compares the excluded path against a current path.

```go
func EqualWithExclusion(a, b *yaml.Node, excludedPath []string) bool {
  return equalWithExclusion(a, b, excludedPath, nil)
}

func equalWithExclusion(a, b *yaml.Node, excluded, current []string) bool { ... }
```

internal/yml/yml_test/compare_test.go
```diff
@@ -132,3 +134,145 @@ foo:
 		})
 	}
 }
+
+func TestEqualWithExclusion(t *testing.T) {
```
davidmdm, line 138: In the land of yaml, we can use flow style syntax, which is kind of a lightweight JSON, which allows us to keep indentation so that the files stay 🥨 (closest emoji to "pretty" was "pretzel")

```go
			yaml1: `
foo:
  bar: baz
  nested:
    - item1
    - item2
`,
```

Becomes (or even nicer because I am struggling with github formatting in the rich editor):

```go
			yaml1: `{
				foo: {
				   bar: baz,
				   nested: [item1, item2],
				},
			}`
```

internal/yml/yml_test/compare_test.go
```diff
@@ -132,3 +134,145 @@ foo:
...
+  bar: baz
+`,
+			yaml2: `
+foo:
+  bar: bam
+`,
+			excludedPath: []string{"foo", "nested"},
+			expected:     false,
+		},
+	}
```
davidmdm, line 261: I would add a test case for the bug (Get ready for more Xmas PLOUP)

```go
{
	name: "ploup!",
	yaml1: `{
		a: {
			b: alpha
		}
	}`,
	yaml2: `{
		a: {
			b: beta
		}
	}`,
	excludedPath: []string{"definitely-not-a", "b"},
	expected:     false,
},
```

You'll notice that the objects are not equal, but since the values that differentiate are at the same level as the expected key and have the same last segment their difference gets ignored

davidmdm (changes requested):
There are some comments that are mostly matters of opinion and taste, so although I left them for good measure, non of them are blocking and I am happy to approve the PR without those changes.

However the `EqualWithExclusion` func has a bug and that's the only blocker!

internal/release/cross/release.go
davidmdm: It would have looked more like `ploup` branch where the we check if values are in sync when computing the list, rather than in a precompute stage.

internal/release/promote/interactive_prompt_provider.go
davidmdm: Its ok! I find that `GetReleaseDisplayVersion(item.Releases[1], true, true, true)` is a little inscrutable. However if you like this, we can leave it as is!

davidmdm (changes requested):
I love the changes that you made around styling. Although you didn't decouple content from style, it made each path much more explicit and easy to understand.

The compare func no longer has the bug it used to have, but I would spend 10 minutes pairing on it together to just clean it up and add support for sequences.

davidmdm (approved):
BUENO

PR #67, opened 2023-12-23
fix(PL-1906): gha, fix handling of BREAKING

j-martin:
Note that the action is the same code as nestoca/conventional-changelog-action@releases/v4, but we this way nobody can change the code and tag under our feet.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -68,8 +68,9 @@ jobs:
 
       - name: Generate changelog and tag release
         id: changelog
-        uses: TriPSs/conventional-changelog-action@v4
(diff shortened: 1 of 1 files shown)


alexstojda (approved):

davidmdm:
Do we really want to maintain a fork? Given that it turns out we don't need to patch anything within the action? 
I understand we get some stability guarantees from forking it and referencing a version we control, but we also forgo maintenance from the authors (until we check and sync our fork manually)

j-martin:
It's a supply chain concern. By referencing a (mutable) tag from a repo, we lose the ability to validate upstream code changes, know when things have changed, or know if the author suddenly made the repro private or deleted.

PR #68, opened 2024-01-19
feat(PL-2250): add --auto-merge flag for joy release promote

davidmdm:


internal/release/promote/interactive_prompt_provider.go
```diff
@@ -182,6 +178,11 @@ func (i *InteractivePromptProvider) ConfirmCreatingPromotionPullRequest() (bool,
 	return selectedAction == actions[0], nil
 }
 
+func (*InteractivePromptProvider) ConfirmAutoMergePullRequest() (answer bool, err error) {
+	err = survey.AskOne(&survey.Confirm{Message: "Do you wish to auto-merge the resulting PR?"}, &answer)
```
j-martin, line 182: Do we ever show a diff before creating the PR?

davidmdm: Yes the diff is shown just before:
image

j-martin: Ah... yes...

internal/release/promote/promotion.go
```diff
@@ -129,10 +131,22 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
 		return "", nil
 	}
 
-	prURL, err := p.perform(list)
+	if opts.TargetEnv.Spec.Promotion.FromPullRequests && !opts.AutoMerge {
```
silphid, line 134: I don't think you lend the proper semantic to `FromPullRequests` property. It should probably be properly commented and documented to make things clearer, but what it really controls is whether `joy pr promote` will consider that environment in the user-prompted list as a potential candidate for promoting a "source code" PR. We will need something else to configure `auto-merge`. Maybe a new `Environment.spec.promotion.autoMergeable` or `.allowAutoMerge` property?

silphid (changes requested):
See my comment but I don't think you interpret the meaning of `TargetEnv.Spec.Promotion.FromPullRequests` correctly.

silphid (approved):

PR #73, opened 2024-02-01
feat(PL-2314): add joy release render command

davidmdm:


cmd/joy/release.go
```diff
@@ -181,3 +193,179 @@ This command requires the jac cli: https://github.com/nestoca/jac
...
+	if err := tmpl.Execute(&builder, params); err != nil {
+		return nil, err
+	}
+
+	var result map[string]any
+	if err := yaml.Unmarshal(builder.Bytes(), &result); err != nil {
+		return nil, err
+	}
+
+	return result, nil
```
j-martin, line 370: When generating the value, we need to inject the `image.tag` value from the version field in the joy spec.

https://github.com/nestoca/infra/blob/7d7daedc210199d5539d009431e94f1dc7b7ff36/tools/joy-converter/joy_converter.py#L439

internal/config/config.go
```diff
@@ -14,6 +14,7 @@ import (
 const (
 	joyrcFile     = ".joyrc"
 	joyDefaultDir = ".joy"
+	joyCacheDir   = ".joycache"
```
j-martin, line 17: I don't think we should clobber the home dir more than we already do.

We should read `XDG_CACHE_HOME` and default to `~/.cache/joy` if it is not defined.

On macOS it would be more correct to be `~/Library/Caches/joy`, but `~/.cache` should be  good enough.

internal/helm/helm.go
```diff
@@ -0,0 +1,59 @@
...
+	return cmd.Run()
+}
+
+func Render(ctx context.Context, w io.Writer, chartPath string, values map[string]any) error {
+	var input bytes.Buffer
+	if err := yaml.NewEncoder(&input).Encode(values); err != nil {
+		return err
+	}
+
+	cmd := exec.CommandContext(ctx, "helm", "template", "joy-release-render", chartPath, "--values", "-")
```
j-martin, line 54: For my education, what's the benefit of carrying the context here?

j-martin (commented):
Cool stuff!

internal/helm/helm.go
davidmdm: In this particular case, none - but only because we haven't added signal cancelation to our root context yet.
When we do, when we hit ctl+c instead of forcefully terminating the program, we can cancel our context and give the program a chance to:
- cleanup whatever its doing.
  - finish writing its file
  - close a network connection
- More importantly - return an error about what happened when it was canceled
  - failed to render chart: context canceled: interrupt
  - failed to pull chart: context canceled: interrupt

As a general rule of good practice, in Go, its important to not shy away from contexts.

cmd/joy/release.go
davidmdm: If I do this it breaks certain chart renders like redis:
```
Error: YAML parse error on redis/templates/master/application.yaml: error converting YAML to JSON: yaml: line 70: mapping values are not allowed in this context
```

I also searched the catalog and never see `image.tag` except for in `podinfo` as the spec.versionKey.
Very confused image.tag

internal/release/render/render.go
```diff
@@ -0,0 +1,178 @@
...
+	CacheDir     string
+	Catalog      *catalog.Catalog
+	IO           internal.IO
+	Helm         helm.PullRenderer
+}
+
+func Render(ctx context.Context, params RenderParams) error {
+	environment, err := getEnvironment(params.Catalog.Environments, params.Env)
+	if err != nil {
+		return fmt.Errorf("failed to get environment: %w", err)
```
silphid, line 36: When wrapping error, let's use positive format:
```suggestion
		return fmt.Errorf("getting environment: %w", err)
```
Same thing for all other errors.

silphid (changes requested):
Magnifique!! 🚀👍  Juste quelques détails mineurs!

cmd/joy/release.go
j-martin: Yah, it's because it's specific to that chart. I am not sure if it is worth handling and just assume that the `image.tag` will be rendered incorrectly.

internal/release/render/render.go
davidmdm: I find this positive format wrapping to be unclear. Its more terse and reads less nice, and you need to have some top level initial first message be the "negative" message but its hard to say where that would happen.

If you look at the errors in the tests they are clear such as:
```
failed to get environment: not found: quality-assurance
failed to get release: not found within environment qa: app
```

I think the positive wrap will make them unclear. Like a sentance without a beginning.

silphid: We already discussed that point in length before your coming and settled on that style, so please let's keep on sticking to it. FYI, the idea is that when you have deeply nested error messages, they become very redundant and verbose, with no added value.  Usually, errors are already displayed with "error: ...", and we would get something like "error: failed to load template file: failed to parse line 5: failed to parse unexpected character ':' in token 'asdf:asdf'", but rather have the more straight to the point version: "error: loading template file: parsing line 5: unexpected character ':' in token 'asdf:asdf'"

internal/helm/helm.go
j-martin: Yep, I should have read the docs:

```
package main

import (
	"context"
	"os/exec"
	"time"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
	defer cancel()

	if err := exec.CommandContext(ctx, "sleep", "5").Run(); err != nil {
		// This will fail after 100 milliseconds. The 5 second sleep
		// will be interrupted.
	}
}

```

internal/release/render/render.go
davidmdm: Will change it

silphid (approved):

PR #79, opened 2024-02-09
fix/PL 2314/replicate appset behaviour in release render

davidmdm:
This replicates the ApplicationSet defaults in release rendering. The only omission is `release.spec.versionKey` because it is not captured in the v1alpha1.Release definition. However nobody uses it the catalog so its ok to omit for now.

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -227,14 +227,15 @@ func NewReleaseRenderCmd() *cobra.Command {
 			}
 
 			return render.Render(cmd.Context(), render.RenderOpts{
-				Env:          env,
-				Release:      releaseName,
-				DefaultChart: cfg.DefaultChart,
-				CacheDir:     cfg.JoyCache,
-				Catalog:      cat,
-				IO:           io,
-				Helm:         helm.CLI{IO: io},
-				Color:        color,
+				Env:           env,
+				Release:       releaseName,
+				DefaultChart:  cfg.DefaultChart,
+				CacheDir:      cfg.JoyCache,
+				ChartMappings: cfg.ChartMappings,
+				Catalog:       cat,
+				IO:            io,
+				Helm:          helm.CLI{IO: io},
+				Color:         color,
 			})
 		},
 	}
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -227,15 +227,15 @@ func NewReleaseRenderCmd() *cobra.Command {
 			}
 
 			return render.Render(cmd.Context(), render.RenderOpts{
-				Env:           env,
-				Release:       releaseName,
-				DefaultChart:  cfg.DefaultChart,
-				CacheDir:      cfg.JoyCache,
-				ChartMappings: cfg.ChartMappings,
-				Catalog:       cat,
-				IO:            io,
-				Helm:          helm.CLI{IO: io},
-				Color:         color,
+				Env:          env,
+				Release:      releaseName,
+				DefaultChart: cfg.DefaultChart,
+				CacheDir:     cfg.JoyCache,
+				ValueMapping: cfg.ValueMapping,
+				Catalog:      cat,
+				IO:           io,
+				Helm:         helm.CLI{IO: io},
+				Color:        color,
 			})
 		},
 	}
--- a/internal/release/render/render.go
+++ b/internal/release/render/render.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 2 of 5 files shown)


riclima (approved):

internal/release/render/render.go
```diff
@@ -162,6 +163,13 @@ func getReleaseViaPrompt(releases []*cross.Release, env string) (*v1alpha1.Relea
 func hydrateValues(release *v1alpha1.Release, environment *v1alpha1.Environment) (map[string]any, error) {
 	values := release.Spec.Values
 
+	// --- This replicates the behavior we have encoded into the application set.
+	if release.Spec.Version != "" {
+		setInMap(values, []string{"image", "tag"}, release.Spec.Version)
+	}
+	setInMap(values, []string{"common", "annotations", "nesto.ca/deployed-by"}, "joy")
+	// --- end of appset behaviour
+
```
alexstojda, line 172: Hmmmm, this is starting to drift into having nesto specific logic in joy, which i'm not sure that we want...

davidmdm: This is true. But this is because we've coupled our appset to our generic chart and we cannot do release rendering in a sane way without knowing these.

The simplest would have been to not have the release.spec.version coupled to the image tag.
Also it would have been good to not have the appset muddy the values of the charts with defaults from the release spec.

That being said, I have been using this command more and more as I work with joy releases. It's actually great when you want to quickly know what a change to a release would produce. The real source of truth is always in argo/k8 but i think getting this to work correctly. Is important especially as we migrate towards joy internally.

My vote is that we allow this much internal logic to bleed into the Joy CLI for now, and work to decouple some stuff so that we can remove this.

@silphid what are your thoughts?

davidmdm: As a thought, if a user from outside nesto opens an issue because the release render applies `image.tag` to their chart values in an undesirable way, then we can address it. However the likelihood of that happening is near zero.

silphid: For `common.annotations.nesto\.ca/deployed\-by: joy` we should do same as for defaulting chart to `generic`, by introducing a catalog-level config for specifying a list of global custom chart parameters. For example:
```
globalChartParams:
  common.annotations.nesto\.ca/deployed\-by: joy
```
(my understanding is that the backslashes are required by helm to properly escape the `nesto.ca/deployed-by` annotation key)

For `image.tag` it's slightly different, as it's just the default version key to use if not specified in release. So I'd simply create a `defaultVersionKey` catalog-level config that we would set to `image.tag` in our case.

Note: Eventually, we should also have the generator use those configs dynamically instead of hard-coding them in the app set, so we have a single source of truth that is used cohesively by both CLI and generator.

silphid (approved):

silphid (changes requested):
See suggestion for keeping joy generic...

cmd/joy/release.go
```diff
@@ -227,14 +227,15 @@ func NewReleaseRenderCmd() *cobra.Command {
...
-				CacheDir:     cfg.JoyCache,
-				Catalog:      cat,
-				IO:           io,
-				Helm:         helm.CLI{IO: io},
-				Color:        color,
+				Env:           env,
+				Release:       releaseName,
+				DefaultChart:  cfg.DefaultChart,
+				CacheDir:      cfg.JoyCache,
+				ChartMappings: cfg.ChartMappings,
```
silphid, line 234: Just unsure about the name... That _could_ work, though I would probably prefer something that emphasizes the values more than the chart, like `ValuesMapping` (and singular form for "Mapping"). But, it can be up for debate! ;)

silphid (approved):
Great, just a few details to maybe consider/discuss together...

internal/release/render/render.go
```diff
@@ -215,3 +221,63 @@ func (w ManifestColorWriter) Write(data []byte) (int, error) {
 	n, err := w.dst.Write([]byte(strings.Join(lines, "\n")))
 	return min(n, len(data)), err
 }
+
+func setInMap(mapping map[string]any, segments []string, value any) {
+	for i, key := range segments {
+		if i == len(segments)-1 {
+			if _, ok := mapping[key]; !ok {
```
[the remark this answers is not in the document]
davidmdm: They are defined by the tests! But I can add a comment.

cmd/joy/release.go
davidmdm: The name is hard to pin down. Its an arbitrary patching of values as defined by joy catalog configuration.

I will go with ValueMapping

internal/release/render/render.go
```diff
@@ -215,3 +221,63 @@ func (w ManifestColorWriter) Write(data []byte) (int, error) {
...
+
+		submap, ok := subValue.(map[string]any)
+		if !ok {
+			return
+		}
+		mapping = submap
+	}
+}
+
+func splitIntoPathSegments(input string) (result []string) {
```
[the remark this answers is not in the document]
davidmdm: The goal of this function was not optimization but correctness. Adding reserved strings/markers makes this more brittle and introduces invalid inputs. Open to alternatives but unless this function is hard to understand I suggest we leave as is.

PR #80, opened 2024-02-09
chore(PL-2239): add draft pull request

ayahajar:


internal/release/promote/interactive_prompt_provider.go
```diff
@@ -164,18 +168,20 @@ func alignColumns(lines []string) []string {
 	return strings.Split(result, "\n")
 }
 
-func (i *InteractivePromptProvider) ConfirmCreatingPromotionPullRequest() (bool, error) {
-	actions := []string{"Create pull request", "Cancel"}
+func (i *InteractivePromptProvider) SelectCreatingPromotionPullRequest() (answer string, err error) {
+	var selectedAction string
+	actions := []string{Draft, Ready, Cancel}
```
davidmdm, line 173: let's make Ready the first option! I think it should be the most used, so for least annoyance it should be easiest to reach

internal/release/promote/promotion.go
```diff
@@ -126,27 +126,34 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
...
+	case Draft:
+		performParams.draft = true
+		break
+	case Ready:
+		if opts.TargetEnv.Spec.Promotion.AllowAutoMerge && !opts.AutoMerge {
+			autoMerge, err := p.promptProvider.ConfirmAutoMergePullRequest()
+			if err != nil {
+				return "", fmt.Errorf("confirming automerge: %w", err)
+			}
+			opts.AutoMerge = autoMerge
```
davidmdm, line 147: here you do not need to modify opts.AutoMerge since you are modifying performParams.autoMerge.

I modified opts before, so that I can use that value when declaring the PerformParams.

davidmdm (changes requested):
This is looking great! All thats needed is:
- some basic tests
- a flag for the command so that it can be pre-selected at the command line
  - joy release promote --draft

internal/release/promote/interactive_prompt_provider.go
ayahajar: Changed

internal/release/promote/promotion.go
```diff
@@ -214,3 +209,28 @@ func getTargetEnvironments(environments []*v1alpha1.Environment, sourceEnvironme
 	}
 	return envs, nil
 }
+
+func (p *Promotion) selectCreatingPR(opts *Opts) (bool, error) {
```
davidmdm, line 213: I am not sure I like the division of this function.
What I don't like:
- It does the prompting, and modifies the opts pointer. Modifying the opts from this func might be unexpected to the caller.
- It introduces a cancel flag which is unneeded if you don't break this function out.

davidmdm (approved):
Small comment but nothing that blocks approval

davidmdm (approved):

PR #87, opened 2024-02-16
feat/PL-2409/add-release-git-diff-and-log

davidmdm:
add joy release git diff and log**

git


cmd/joy/release.go
```diff
@@ -258,3 +266,111 @@ func NewReleaseRenderCmd() *cobra.Command {
...
+
+			getRevision := func(version string) string {
+				version = "v" + version
+				if branch := semver.Prerelease(version); branch != "" {
+					return branch
+				}
+				return path.Join(project.Spec.RepositorySubdirectory, version)
+			}
+
+			pull := exec.Command("git", "pull")
```
j-martin, line 350: Instead of a pull, I would do a `git fetch --tags`. That way the state of working tree is inconsequential.

j-martin: Also the `--tags` is important to fetch new tags.

cmd/joy/release.go
```diff
@@ -258,3 +266,111 @@ func NewReleaseRenderCmd() *cobra.Command {
...
+			}
+
+			if sourceRelease == nil {
+				return fmt.Errorf("no source found")
+			}
+			if targetRelease == nil {
+				return fmt.Errorf("no target found")
+			}
+
+			repoCache := filepath.Join(cfg.JoyCache, "src")
```
j-martin, line 315: My first thought would have been to use the existing repo, but I agree having them standalone simplifies things.

cmd/joy/release.go
```diff
@@ -258,3 +266,111 @@ func NewReleaseRenderCmd() *cobra.Command {
...
+			pull := exec.Command("git", "pull")
+			pull.Dir = repoDir
+
+			if err := pull.Run(); err != nil {
+				return fmt.Errorf("failed to pull project: %w", err)
+			}
+
+			expr := getRevision(sourceRelease.Spec.Version) + ".." + getRevision(targetRelease.Spec.Version)
+
+			gitCommand := exec.Command("git", command, expr)
```
j-martin, line 359: Can we make it that the user can pass additional flags? e.g.: `-w`?

j-martin: It would also be nice if we output the full command used (and the working dir) so the user can execute it themselves or re-use the commit range for example.

j-martin (approved):
Great!

cmd/joy/release.go
davidmdm: Probably need to do both a `git fetch` plus `git fetch --tags` since we need to be able to pull branches as well for environments that have PRs deployed to them

cmd/joy/release.go
davidmdm: We can't really have people put their local configurations in the catalog, or assume their folder/repo-cloning conventions. So yeah! Disk space is free! maybe?

cmd/joy/release.go
j-martin: Only `git fetch --tags` is required. It fetches more than just tags. That being said, we only care about tags.

cmd/joy/release.go
davidmdm: to the prior, maybe! It being arbitrary flags might be hard with cobra. It's something I could do (and do do) with the standard flag package all the time, but yeah.

cmd/joy/release.go
davidmdm: 👍

cmd/joy/release.go
davidmdm: found a way possibly

cmd/joy/release.go
```diff
@@ -258,3 +267,117 @@ func NewReleaseRenderCmd() *cobra.Command {
...
 	return cmd
 }
+
+func NewGitQueryCommand(command string) *cobra.Command {
+	var (
+		source string
+		target string
+	)
+	cmd := &cobra.Command{
+		Use:  "git-" + command,
```
j-martin, line 277: Can we have `git diff` instead? It's more a matter of taste, though.

api/v1alpha1/project.go
```diff
@@ -31,6 +31,10 @@ type ProjectSpec struct {
 
 	// Location of the project in the repository. Should be empty if the whole repository is the project.
 	RepositorySubdirectory string `yaml:"repositorySubdirectory,omitempty" json:"repositorySubdirectory,omitempty"`
+
+	// GitTagTemplate allows you to configure what your git tag look like relative a to a release via go templates
+	// example: gitTagTemplate: api/v{{ .Release.Spec.Versiion }}
```
davidmdm, line 36: typo

silphid (approved):

PR #88, opened 2024-02-17
feat: add support for catalog.yaml

j-martin:
Somewhat of a proposal, I think having an hidden file `.joyrc` in the `catalog` is a great idea, but I think it should be visible (easily) to the user so I am proposing we rename it to `catalog.yaml`.

Any other name is welcome.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	CatalogConfigFile = "catalog.yaml"
+	CatalogConfigFile = "joy.yaml"
 	JoyrcFile         = ".joyrc"
 	JoyDefaultDir     = ".joy"
 )
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	JoyrcFile     = ".joyrc"
-	JoyDefaultDir = ".joy"
+	CatalogConfigFile = "catalog.yaml"
+	JoyrcFile         = ".joyrc"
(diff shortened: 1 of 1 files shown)


davidmdm:
I am OK with these changes, however they need to be matched in the joy CLI before we can merge this otherwise we will break existing commands/functionality.

Therefore:
- add support catalog.yaml and .joyrc to the joy-cli
- change the minimumVersion in the catalog to the version that supports catalog.yaml
- wait a reasonable amount of time for folks to update.
- rename the .joyrc file

Suggestion: it seems to me that having a `catalog.yaml` in a `catalog` repo is a little repetitive. How about `joy.yaml`?

alexstojda (approved):

davidmdm (changes requested):
Setting a request changes in order to make sure we don't merge this accidentally. Joy needs to be updated first.

j-martin:
> add support catalog.yaml and .joyrc to the joy-cli

That's what this PR does.

> Suggestion: it seems to me that having a catalog.yaml in a catalog repo is a little repetitive. How about joy.yaml?

I have renamed the file to `joy.yaml`. I also have tested using a symlink and it was handled gracefully.

So we can just do:

```
git mv .joyrc joy.yaml
ln -s joy.yaml .joyrc
git add .
```

And delete the symlink in a few months or so.

j-martin:
> This should be a feat or a fix. A chore does not bump the application version.

Yah, I'll have to re-open a new PR.

j-martin:
Closing in favor of https://github.com/nestoca/joy/pull/92

PR #89, opened 2024-02-20
fix/PL-2372: add release ignorelist for value mappings

davidmdm:
- **fix(PL-2372): add release ignore list for value mappings**

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -46,12 +46,38 @@ type Config struct {
 	//   image.tag: {{ .Release.Spec.Version }}
 	//   common.annotations.example\.com/custom: true
 	//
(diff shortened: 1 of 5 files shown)


davidmdm:
Currently:
```
go run ./cmd/joy --skip-version-check release render -e qa m2mtokens-redis
Error: YAML parse error on redis/templates/master/application.yaml: error converting YAML to JSON: yaml: line 70: mapping values are not allowed in this context

Use --debug flag to render out invalid YAML
Error: rendering chart: exit status 1
exit status 1
```

With an update to the catalog:
```
valueMapping:
  releaseIgnoreList:
  - m2mtokens-redis
  mappings:
    image.tag: '{{ .Release.Spec.Version }}'
    common.annotations.nesto\.ca/deployed-by: joy
```

We have success.

PR #90, opened 2024-02-20
feat/PL-2409 - Infer repo name

j-martin:
Instead of having to set it everywhere, we can infer it but still allow it to be overridden.

### DIFF
--- a/cmd/joy/root.go
+++ b/cmd/joy/root.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


j-martin:
This is PR against the `feat/PL-2409/add-release-git-diff-and-log` branch and not `master`.

cmd/joy/root.go
```diff
@@ -58,7 +59,7 @@ func NewRootCmd(version string) *cobra.Command {
 				if version != "" && semver.Compare(version, cfg.MinVersion) < 0 {
 					return fmt.Errorf("current version %q is less than required minimum version %q. Please update joy", version, cfg.MinVersion)
 				}
-				if !semver.IsValid(version) {
+				if !semver.IsValid(version) && os.Getenv("JOY_DEV_NO_PROMPT") != "1" {
```
davidmdm, line 62: you don't need this. You can use the secret `--skip-version-check` for development builds

davidmdm (approved):
Approve with 1 comment

cmd/joy/root.go
j-martin: I moved and renamed the env var to match that.

PR #92, opened 2024-02-20
feat(PL-2419): add support for alternate names for catalog/joy.yaml

j-martin:
I have renamed the file to `joy.yaml` as discussed in https://github.com/nestoca/joy/pull/88

I also have tested using a symlink and it was handled gracefully
by `joy` as-is.

So we can just do: https://github.com/nestoca/catalog/pull/597

And delete the symlink in a few months or so.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	JoyrcFile     = ".joyrc"
-	JoyDefaultDir = ".joy"
+	CatalogConfigFile = "joy.yaml"
+	JoyrcFile         = ".joyrc"
+	JoyDefaultDir     = ".joy"
 )
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


riclima (approved):

silphid (changes requested):
Yeah, agreeing with the symlink approach! 👍  Let's just amend the code to only support the new file name, instead of keeping backward compatibility!

j-martin:
@silphid GH won't let me re-request a review from you, but can you have another look?

image

silphid (approved):

internal/config/config.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	JoyrcFile     = ".joyrc"
-	JoyDefaultDir = ".joy"
+	CatalogConfigFile = "joy.yaml"
```
davidmdm, line 15: I think we need to be delicate here since the catalog repo now controls the joy-generator, and if we don't update the generator to a compatible version this change could break things with argocd.

I would update this such that the catalogConfigFile is a list of candidates: .joyrc and joy.yaml, such that it looks for joy.yaml first and falls back to .joyrc.

Then we can update the joy-generator, then we can update the catalog.

j-martin: I have merged https://github.com/nestoca/catalog/pull/597 yesterday and joy/joy generator have been working the same.

silphid: @davidmdm the symlink does the backward compatibility magic here.

PR #95, opened 2024-02-21
feat: add aliases to some commands

j-martin:


riclima (approved):

silphid (approved):
Those are some quite radical shorthand aliases! 😆

j-martin:
Yes, then you can alias joy and do stuff like this: `j r g d`

PR #97, opened 2024-02-21
fix: tell people to run `brew update` first

j-martin:
Otherwise we depend on the state of brew's cache.

### DIFF
--- a/internal/diagnostics/executable.go
+++ b/internal/diagnostics/executable.go
@@ -39,7 +39,7 @@ func diagnoseExecutable(cfg *config.Config, cliVersion string, opts ExecutableOp
 		group.AddMsg(
 			failed,
 			fmt.Sprintf("Version does not meet minimum of %s required by catalog", style.Code(cfg.MinVersion)),
-			msg(hint, fmt.Sprintf("Update joy using: %s", style.Code("brew upgrade joy"))),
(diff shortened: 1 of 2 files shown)


riclima (approved):

alexstojda (approved):

davidmdm (approved):

davidmdm:
@alexstojda that was a misclick. Ignore the rerequest for review. Its not even my PR.

internal/diagnostics/executable.go
```diff
@@ -39,7 +39,7 @@ func diagnoseExecutable(cfg *config.Config, cliVersion string, opts ExecutableOp
 		group.AddMsg(
 			failed,
 			fmt.Sprintf("Version does not meet minimum of %s required by catalog", style.Code(cfg.MinVersion)),
-			msg(hint, fmt.Sprintf("Update joy using: %s", style.Code("brew upgrade joy"))),
+			msg(hint, fmt.Sprintf("Update joy using: %s", style.Code("brew update && brew upgrade joy"))),
```
silphid, line 42: @alexstojda I thought you mentioned `brew update` was no longer required, so I updated the different READMEs accordingly, but I can't find any docs supporting that... 🤔

alexstojda: Yeah I thought it was the case, and do recall seeing it somewhere, but I too can no longer find docs supporting it. 

Perhaps it's not the case for 3rd party taps?

PR #98, opened 2024-02-22
fix(PL-2424): filter git diff/logs correctly in mono-repos

j-martin:
Note that the repositorySubdirectory field was not used.

https://github.com/nestoca/catalog/pull/692 must be merged after.

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/TwiN/go-color"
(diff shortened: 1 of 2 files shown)


cmd/joy/release.go
```diff
@@ -384,17 +385,25 @@ func NewGitCommands() *cobra.Command {
...
+				fmt.Println(color.InCyan("running:"), color.InBold(strings.Join(gitArgs, " ")), color.InCyan("in"), color.InBold(repoDir))
 				fmt.Println()
 
-				gitCommand := exec.Command("git", gitargs[1:]...)
+				gitCommand := exec.Command("git", gitArgs[1:]...)
 				gitCommand.Dir = repoDir
 				gitCommand.Stdout = os.Stdout
 				gitCommand.Stderr = os.Stderr
 				gitCommand.Stdin = os.Stdin
+				gitCommand.Env = os.Environ()
```
j-martin, line 406: Otherwise, I can't hijack stuff like the `GIT_PAGER.`

davidmdm: Are you sure? If its not present it defaults to os.Environ:
```
Env specifies the environment of the process. Each entry is of the form "key=value". If Env is nil, the new process uses the current process's environment
```

cmd/joy/release.go
```diff
@@ -384,17 +385,25 @@ func NewGitCommands() *cobra.Command {
...
-				gitargs = append(gitargs, sourceTag+".."+targetTag)
+				gitArgs := append([]string{"git", command}, args[1:]...)
+				gitArgs = append(gitArgs, sourceTag+".."+targetTag)
 
-				fmt.Println(color.InCyan("running:"), color.InBold(strings.Join(gitargs, " ")))
+				if project.Spec.RepositorySubpaths != nil {
+					if !slices.Contains(gitArgs, "--") {
+						gitArgs = append(gitArgs, "--")
+					}
+					gitArgs = append(gitArgs, project.Spec.RepositorySubpaths...)
```
davidmdm, line 395: Why do we want a project to have multiple subpaths?

Doesn't it it make for sense to have multiple projects with the same repo but different subpaths?

IE:
```
project: applicartions -> nestoca/public-apps + /applications
project: documents    -> nestoca/pulbic-apps + /documents
...
```

This way we only log the changes related to the subpath related to the project?

I may be misunderstanding either how are projects are setup or how git log -- paths... work?

j-martin: Projects changes are much more than just the `app/<app-name>`

See https://github.com/nestoca/catalog/pull/692/files?w=1#diff-997462567ee0af560610519be390296da8a66b6907d044d59a4f78962287c2dbR10-R13

davidmdm (approved):

PR #99, opened 2024-02-22
fix(PL-2422): don't allow branch promotion

ayahajar:


internal/release/cross/release_list.go
```diff
@@ -147,6 +149,15 @@ func (r *ReleaseList) GetReleasesForPromotion(sourceEnv, targetEnv *v1alpha1.Env
...
 		sourceRelease := item.Releases[sourceEnvIndex]
 		targetRelease := item.Releases[targetEnvIndex]
+
+		//Check for version in source Release
+		if targetEnv.Name == "qa" || targetEnv.Name == "production" {
+			version := "v" + sourceRelease.Spec.Version
+			if semver.Prerelease(version) != "" || semver.Build(version) != "" {
+				continue
+			}
+		}
```
silphid, line 159: We should never hard-code any specific environment names in Joy, those are nesto specific, and even if Joy was not open source we should never hard-code that kind of logic! 😅  As we discussed yesterday, the check should use the environment's promotion configuration:
```suggestion
		version := "v" + sourceRelease.Spec.Version
		if !targetEnv.Spec.Promotion.FromPullRequests && (semver.Prerelease(version) != "" || semver.Build(version) != "") {
			continue
		}
```

silphid (changes requested):
See my comment...

internal/release/cross/release_list.go
```diff
@@ -151,7 +151,7 @@ func (r *ReleaseList) GetReleasesForPromotion(sourceEnv, targetEnv *v1alpha1.Env
 		targetRelease := item.Releases[targetEnvIndex]
 
 		//Check for version in source Release
-		if targetEnv.Name == "qa" || targetEnv.Name == "production" {
+		if !targetEnv.Spec.Promotion.FromPullRequests {
```
davidmdm, line 154: I don't think we want to do filtering as previously discussed!

internal/release/cross/release_list.go
```diff
@@ -171,6 +171,24 @@ func (r *ReleaseList) GetReleasesForPromotion(sourceEnv, targetEnv *v1alpha1.Env
 	return subset, nil
 }
 
+// HasNonPromotableReleases checks if the list contains releases that cannot be promoted based
+// on the version format allowed at the target environment.
+func (r *ReleaseList) HasNonPromotableReleases(sourceEnv, targetEnv *v1alpha1.Environment) bool {
```
davidmdm, line 176: I would rename the function to something more indicative of what it is checking for. `GetInvalidReleaseVersions`.

I also think it should be a `get` and not a `has` because it would be nice if the error could include the release names that were invalid:

Something like:
```
cannot promote release(s): audittrail, products, external-brokerage: target env qa does not allow pre-release versions 
```

internal/release/cross/release_list.go
```diff
@@ -163,9 +163,10 @@ func (r *ReleaseList) GetReleasesForPromotion(sourceEnv, targetEnv *v1alpha1.Env
 	return subset, nil
 }
 
-// HasNonPromotableReleases checks if the list contains releases that cannot be promoted based
+// GetInvalidReleaseVersions returns a list of names of releases that cannot be promoted based
 // on the version format allowed at the target environment.
-func (r *ReleaseList) HasNonPromotableReleases(sourceEnv, targetEnv *v1alpha1.Environment) bool {
+func (r *ReleaseList) GetInvalidReleaseVersions(sourceEnv, targetEnv *v1alpha1.Environment) []string {
```
silphid, line 168: Haha, I know it's David who proposed that name, but I disagree, because those are not invalid, they are just pre-release. I would opt for something more aligned with the real semantic/intent of the function, very similar to what you had initially (which was good, though David suggested to get the specific list) so something like `GetNonPromotableReleases`.

silphid (approved):
I will pre-approve, to avoid too many review cycles, but just consider my comment and go ahead with the merge! 👍

internal/build/promote.go
```diff
@@ -29,12 +31,21 @@ func Promote(opts Opts) error {
...
 		release := crossRelease.Releases[0]
 		if release.Spec.Project == opts.Project {
+
 			// Find version node
 			versionNode, err := yml.FindNode(release.File.Tree, "spec.version")
 			if err != nil {
 				return fmt.Errorf("release %s has no version property: %w", release.Name, err)
 			}
 
+			// Check the version format in source Release
```
davidmdm, line 41: I think that this would be clearer if it wasn't done inside of the loop. What we need to check is that version is valid or not for the environment. I think you could do this before entering the loop:

```go
isPr := func() bool {
	version := "v" + opts.Version
	return semver.Prerelease(version)+semver.Build(version) != ""
}()

if isPr && !cat.Environments[0].Spec.Promotion.FromPullRequests {
	return fmt.Errorf("cannot promote release with non-standard version to %s environment", opts.Environment)
}
```

internal/build/promote_test.go
```diff
@@ -58,11 +58,28 @@ func TestPromote(t *testing.T) {
...
+		Environment: "testing",
+		Project:     "promote-build",
+		Version:     "1.1.2-updated",
+	}
+
+	_, err := setup(setupOpts{
+		env:     opts.Environment,
+		project: opts.Project,
+	})
+	assert.Nil(t, err)
```
davidmdm, line 72: Prefer `require` over `assert`.

`require` stops the test if the condition is not met. `assert` records the error but continues.
For conditions that render the test invalid such as failed setup -> nothing can go right after that.

Even in conditions where assert is justified, the noise it creates is generally a headache.

internal/release/cross/release_list.go
```diff
@@ -160,6 +163,25 @@ func (r *ReleaseList) GetReleasesForPromotion(sourceEnv, targetEnv *v1alpha1.Env
...
+// GetNonPromotableReleases returns a list of names of releases that cannot be promoted based
+// on the version format allowed at the target environment.
+func (r *ReleaseList) GetNonPromotableReleases(sourceEnv, targetEnv *v1alpha1.Environment) []string {
+	var invalidList []string
+	sourceEnvIndex := r.getEnvironmentIndex(sourceEnv.Name)
+
+	for _, item := range r.Items {
+		sourceRelease := item.Releases[sourceEnvIndex]
+		// Check the version format in source Release
+		if !targetEnv.Spec.Promotion.FromPullRequests {
```
davidmdm, line 175: It would be clearer to the reader if you can perform this check outside of the loop. If the environment accepts PullRequests you don't need to perform this loop.

```go
if sourceEnv.Spec.Promotion.FromPullRequests { 
  return nil
}
```
```

File: internal/release/promote/promotion.go [hunk 8]
```diff
@@ -131,20 +133,23 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
...
 		return "", fmt.Errorf("selecting releases to promote: %w", err)
 	}
-	if !list.HasAnyPromotableReleases() {
-		p.promptProvider.PrintNoPromotableReleasesFound(opts.ReleasesFiltered, opts.SourceEnv, opts.TargetEnv)
-		return "", nil
-	}
 
 	if !opts.NoPrompt {
-		if err := p.preview(list); err != nil {
+		if err := p.preview(selectedList); err != nil {
```
davidmdm, line 138: Is the preview the `git diff`? If so it might worthwhile to error out before showing the diff.

internal/release/promote/promotion.go
```diff
@@ -131,20 +133,23 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
...
 			return "", fmt.Errorf("previewing: %w", err)
 		}
 	}
 
+	invalidList := selectedList.GetNonPromotableReleases(opts.SourceEnv, opts.TargetEnv)
+	if len(invalidList) != 0 {
+		invalid := strings.Join(invalidList, ", ")
```
davidmdm, line 145: This is not really a comment on your PR because the codebase is like this everywhere, but we should stop printing error messages and returning errors.

**The error should represent the message and then be handled and printed later.**

In this case this actually introduces a very little 🐛 . Since the only way to get here is if the err is nil, and so we would print the error message but then exit successfully.

This means that in CI Workflows build promotion could fail, but the workflow would pass.

internal/release/promote/promotion.go
```diff
@@ -131,20 +131,23 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
...
 		return "", fmt.Errorf("selecting releases to promote: %w", err)
 	}
-	if !list.HasAnyPromotableReleases() {
-		p.promptProvider.PrintNoPromotableReleasesFound(opts.ReleasesFiltered, opts.SourceEnv, opts.TargetEnv)
+
+	invalidList := selectedList.GetNonPromotableReleases(opts.SourceEnv, opts.TargetEnv)
+	if len(invalidList) != 0 {
+		invalid := strings.Join(invalidList, ", ")
+		p.promptProvider.PrintSelectedNonPromotableReleases(invalid, opts.TargetEnv.Name)
 		return "", nil
```
davidmdm, line 139: We should not be returning a nil error. We want the program to fail if invalid releases are selected

PR #107, opened 2024-03-05
fix(PL-2463): ensure catalog is at origin head

ayahajar:
ensures the catalog is at origin head
Changing skip-catalog-update to a global flag 
Moving the check of skipCatalogUpdate to PreRun of the command instead of the specific actions

### DIFF
--- a/internal/config/flags.go
+++ b/internal/config/flags.go
@@ -9,11 +9,11 @@ type GlobalFlags struct {
 
 type flagKey struct{}
 
-func ToFlagContext(parent context.Context, flags *GlobalFlags) context.Context {
+func ToFlagsContext(parent context.Context, flags *GlobalFlags) context.Context {
 	return context.WithValue(parent, flagKey{}, flags)
 }
 
-func FromFlagContext(ctx context.Context) *GlobalFlags {
+func FlagsFromContext(ctx context.Context) *GlobalFlags {
 	cfg, _ := ctx.Value(flagKey{}).(*GlobalFlags)
 	return cfg
 }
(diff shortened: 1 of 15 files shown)


silphid (approved):
Good work! 👍

silphid:
Same comment as @davidmdm, nothing is wired up together! 🤷😅

silphid (dismissed):
Please review your code, feels like it should not even compile(?) or at least not consider `--skip-catalog-update` flag at all...

internal/config/flags.go
```diff
@@ -0,0 +1,19 @@
...
+	SkipCatalogUpdate bool
+}
+
+type flagKey struct{}
+
+func ToFlagContext(parent context.Context, flags *GlobalFlags) context.Context {
+	return context.WithValue(parent, flagKey{}, flags)
+}
+
+func FromFlagContext(ctx context.Context) *GlobalFlags {
```
davidmdm, line 16: Can we rename this to `FlagsFromContext`? I don't know. I know you are using the name I used as an example yesterday, and I am not 100% convinced its a better name, and generally speaking I don't mind naming too much.

Feel free to disregard this comment but if you think it is an improvement hit F2! :)

davidmdm (commented):
My main comment for all of these is that I think we are doing the `catalog-update-check` at the wrong place. This wasn't a choice you were making in this PR, you were just following the existing structure, however now that we have this global flag in the command's context, I think it would be most appropriate to do this as `PreRunE` to the commands that need it, and not as part of the logic of specific actions.

If we made that change, I would be very happy with the improvement! 🚀

davidmdm (approved):

PR #109, opened 2024-03-05
feat(PL-2460): add `--local-only` mode

j-martin:
A better name is more than welcome. This is similar to a `--dry-run` (preview, no PR, etc.), but it writes the changes to the local filesystem.

We joked around calling it `--moist-run`. I need this change for my bulk `!lock`ing. So there it is.

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -90,7 +90,7 @@ func NewReleaseListCmd() *cobra.Command {
 
 func NewReleasePromoteCmd() *cobra.Command {
 	var sourceEnv, targetEnv string
-	var autoMerge, draft, dryRun, skipCatalogUpdate, noPrompt bool
+	var autoMerge, draft, dryRun, localOnly, skipCatalogUpdate, noPrompt bool
 
 	cmd := &cobra.Command{
 		Use:     "promote [flags] [releases]",
@@ -115,8 +115,8 @@ func NewReleasePromoteCmd() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, releases []string) error {
-			if skipCatalogUpdate && !dryRun {
-				return fmt.Errorf("flag --skip-catalog-update requires --dry-run")
+			if skipCatalogUpdate && !(dryRun || localOnly) {
+				return fmt.Errorf("flag --skip-catalog-update requires --dry-run or --local-only")
 			}
 
 			cfg := config.FromContext(cmd.Context())
@@ -157,6 +157,7 @@ func NewReleasePromoteCmd() *cobra.Command {
 				AutoMerge:            autoMerge,
 				Draft:                draft,
 				DryRun:               dryRun,
+				LocalOnly:            localOnly,
 				SkipCatalogUpdate:    skipCatalogUpdate,
 				SelectedEnvironments: selectedEnvironments,
 			}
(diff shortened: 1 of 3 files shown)


silphid (changes requested):
Great feature, though I find the name more confusing than anything (because of the likeliness to unrelated concepts of Read-Only, Read/Write, Write-Only).  It gives no clue as to what it really does.  Possible suggestions would be `--local`, `--local-only`, `--skip-pr`. What do you all think @nestoca/stream-platform?

davidmdm (commented):
Whats' the difference with `--dry-run`? Is it just an alias?

I don't see any custom behavior that differentiates write-only from dry-run?

silphid:
> Whats' the difference with `--dry-run`? Is it just an alias?
> I don't see any custom behavior that differentiates write-only from dry-run?

Haha @davidmdm good catch, I didn't even realize it was doing same as dry-run! I guess it's @j-martin testing our review skills (and I just flunked! 😬)

j-martin:
When using `--dry-run` it actively does not write to disk:

https://github.com/nestoca/joy/blob/9cb9895f48bb03341c282a7a899da0d54cd9eb88/internal/release/promote/perform.go#L109-L115

To be fair it took me a few minutes to re-find it. At first, I thought you were right, about the code doing nothing more than `--dry-run`

davidmdm:
Weird I don't see what you posted in the changes...

davidmdm (approved):

j-martin:
@silphid I have renamed it to `--local-only`.

cmd/joy/release.go
```diff
@@ -115,8 +115,8 @@ func NewReleasePromoteCmd() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, releases []string) error {
-			if skipCatalogUpdate && !dryRun {
-				return fmt.Errorf("flag --skip-catalog-update requires --dry-run")
+			if skipCatalogUpdate && !(dryRun || localOnly) {
+				return fmt.Errorf("flag --skip-catalog-update requires --dry-run or --write-only")
```
silphid, line 119:
```suggestion
				return fmt.Errorf("flag --skip-catalog-update requires --dry-run or --local-only")
```

internal/release/promote/perform.go
```diff
@@ -130,9 +131,18 @@ func (p *Promotion) perform(opts PerformOpts) (string, error) {
...
 	}
 
-	branchName := getBranchName(info)
+	modeName := "normal"
 	if opts.dryRun {
-		fmt.Printf("ℹ️ Dry-run: skipping creation of branch %s\nFiles:\n%s\nCommit message:\n%s\n",
+		modeName = "Dry-Run"
+	}
+	if opts.localOnly {
+		modeName = "Write-Only"
```
silphid, line 139:
```suggestion
		modeName = "Local-Only"
```

internal/release/promote/promotion.go
```diff
@@ -118,6 +121,10 @@ func (p *Promotion) Promote(opts Opts) (string, error) {
 		fmt.Println("ℹ️ Dry-run mode enabled: No changes will be made.")
 	}
 
+	if opts.LocalOnly {
+		fmt.Println("ℹ️ Write-only mode enabled: The local repo will be modified, but not committed. No pull request will be created.")
```
silphid, line 125:
```suggestion
		fmt.Println("ℹ️ Local-only mode enabled: The local repo will be modified, but not committed. No pull request will be created.")
```

silphid (approved):
LGTM, only a few forgotten renames...

> To be fair it took me a few minutes to re-find it. At first, I thought you were right, about the code doing nothing more than --dry-run

Wow, yeah, that was subtle and indirect, but you're right! 👍

PR #111, opened 2024-03-07
fix(PL-2247): handle same version promotion

silphid:
This PR is required before merging that other PR:
https://github.com/nestoca/catalog/pull/1278

### DIFF
--- a/internal/release/promote/pr_rendering.go
+++ b/internal/release/promote/pr_rendering.go
@@ -1,0 +1,181 @@
+package promote
+
+import (
+	"fmt"
+	"strings"
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
+	"github.com/nestoca/joy/api/v1alpha1"
+	"golang.org/x/mod/semver"
+)
+
+type ChangeType string
+
+const (
+	ChangeTypeUpgrade   ChangeType = "Upgrade"
+	ChangeTypeDowngrade ChangeType = "Downgrade"
+	ChangeTypeUpdate    ChangeType = "Update"
+)
+
+type ReleaseInfo struct {
+	Name         string
+	Project      *v1alpha1.Project
+	CodeOwners   []string
+	Repository   string
+	Source       ReleaseWithGitTag
+	Target       ReleaseWithGitTag
+	OlderGitTag  string
+	NewerGitTag  string
+	IsPrerelease bool
+	// IsUpgrade is now deprecated and kept only temporarily while we transition to ChangeType
+	IsUpgrade  bool
+	ChangeType ChangeType
+	Commits    []*CommitInfo
+	Error      error
+}
+
+type PromotionInfo struct {
+	SourceEnvironment *v1alpha1.Environment
+	TargetEnvironment *v1alpha1.Environment
+	Releases          []*ReleaseInfo
+	Error             error
+}
+
+type CommitInfo struct {
+	Sha          string
+	ShortSha     string
+	Author       string
+	GitHubAuthor string
+	Message      string
+	ShortMessage string
+}
+
+func getReleaseInfo(sourceRelease, targetRelease *v1alpha1.Release, opts PerformOpts) (*ReleaseInfo, error) {
+	project := sourceRelease.Project
+	changeType := ChangeTypeUpgrade
+	if targetRelease != nil {
+		comparison := semver.Compare("v"+sourceRelease.Spec.Version, "v"+targetRelease.Spec.Version)
+		if comparison < 0 {
+			changeType = ChangeTypeDowngrade
+		} else if comparison > 0 {
+			changeType = ChangeTypeUpgrade
+		} else {
+			changeType = ChangeTypeUpdate
+		}
+	}
+
+	displayTargetVersion := "(undefined)"
+	if targetRelease != nil {
+		displayTargetVersion = targetRelease.Spec.Version
+	}
+
+	sourceTag, err := opts.getReleaseGitTagFunc(sourceRelease)
+	if err != nil {
+		return nil, fmt.Errorf("getting tag for source version %s of release %s: %w", sourceRelease.Spec.Version, sourceRelease.Name, err)
+	}
+
+	targetTag := sourceTag
+	if targetRelease != nil {
+		targetTag, err = opts.getReleaseGitTagFunc(targetRelease)
+		if err != nil {
+			return nil, fmt.Errorf("getting tag for target version %s of release %s: %w", targetRelease.Spec.Version, targetRelease.Name, err)
+		}
+	}
+
+	olderTag := targetTag
+	newerTag := sourceTag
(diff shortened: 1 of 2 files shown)


internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+		} else if comparison > 0 {
+			changeType = ChangeTypeUpgrade
+		} else {
+			changeType = ChangeTypeUpdate
+		}
+	}
+
+	isPrerelease := false
+	if semver.Build("v"+sourceRelease.Spec.Version) != "" ||
+		semver.Prerelease("v"+sourceRelease.Spec.Version) != "" ||
```
ayahajar, line 79: Replace with 
		if semver.Build(version) + semver.Prerelease(version) != ""

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+		(targetRelease != nil && (semver.Build("v"+targetRelease.Spec.Version) != "" || semver.Prerelease("v"+targetRelease.Spec.Version) != "")) {
+		isPrerelease = true
```
ayahajar, line 81: Same here for target version
if semver.Build(version) + semver.Prerelease(version) != ""

davidmdm: Might as well wrap this into a function `isPrerelease`.

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+	if targetRelease != nil {
+		targetTag, err = opts.getReleaseGitTagFunc(targetRelease)
+		if err != nil {
+			return getAndPrintReleaseInfoWithError("getting tag for target version %s of release %s: %w", targetRelease.Spec.Version, targetRelease.Name, err)
+		}
+	}
+
+	olderTag := targetTag
+	newerTag := sourceTag
+	if changeType == ChangeTypeDowngrade {
```
davidmdm, line 104: instead of keeping track of a change type, can't we just do the check here?

```go
if semver.Compare(sourceTag, releaseTag) == 1 {
  sourceTag, targetTag = targetTag, sourceTag
}
```

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+	"fmt"
+	"strings"
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
+	"github.com/nestoca/joy/api/v1alpha1"
+	"golang.org/x/mod/semver"
+)
+
+type ChangeType string
```
davidmdm, line 13: There's generally no reason to use a semantic type like this. It gives us the feeling of an enum, but its just a lie. Constant strings would be just as good, and introduce less mental overhead.

If you want me to show you how to emulate enums in Go, we can do that offline. But generally its only good across package boundaries like when you are writing a lib for others to consume.

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+	Sha          string
+	ShortSha     string
+	Author       string
+	GitHubAuthor string
+	Message      string
+	ShortMessage string
+}
+
+func getReleaseInfo(sourceRelease, targetRelease *v1alpha1.Release, opts PerformOpts) *ReleaseInfo {
+	getAndPrintReleaseInfoWithError := func(msg string, args ...any) *ReleaseInfo {
```
davidmdm, line 55: I think this function would make more send if it returned an error `getReleaseInfo(...) (*ReleaseInfo, error)`.
You are wrapping the error with information that is available outside the function, and it also should be the responsibility of the caller to handle the error:

```go
info, err := getReleaseInfo(source, target, opts)
if err != nil {
  fmt.Printf("⚠️ failed to get info for release %s: %v\n", source.Name, err)
  info = &ReleaseInfo{Name: source.Name, Error: err}
}
```

It's a great example of when you can handle an error by logging it, and continue on your way. Functions that generate errors shouldn't handle themselves.

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+		if len(shortSha) > 7 {
+			shortSha = shortSha[:7]
+		}
+
+		metadata.Message, err = injectPullRequestLinks(repository, metadata.Message)
+		if err != nil {
+			return getAndPrintReleaseInfoWithError("injecting pull request links: %w", err)
+		}
+
+		shortMessage := metadata.Message
```
davidmdm, line 159: shortMessage, _, _ := strings.Cut(metadata.Message, "\n")

internal/release/promote/pr_rendering.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	codeOwners, err := opts.getCodeOwnersFunc(projectDir)
+	if err != nil {
+		return getAndPrintReleaseInfoWithError("getting GitHub authors: %w", err)
+	}
+	releaseInfo.CodeOwners = codeOwners
+
+	for _, metadata := range commitsMetadata {
+		shortSha := metadata.Sha
```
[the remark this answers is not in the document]
davidmdm: PS I don't actually care about condensing things into one liners. So, there's no actual recommendation to change. Given oneliners can become complicated fast for no added value. But thought I would mention it anyway.

internal/release/promote/pr_rendering.go
silphid: Let me disagree here! 😁 Concatenating the strings just to save a comparison does not gain much in terms of anything, and I would even argue that it makes logic a bit less self-evident. However, I agree with David that encapsulating that within a function would help making everything clearer.

internal/release/promote/pr_rendering.go
silphid: No, `changeType` is intended to be exposed to template, either for it to display it or use for conditional branching (see other PR in catalog).

internal/release/promote/pr_rendering.go
silphid: As I mentioned above, change type is exposed to template, so I think it's a good thing to formalize it, along with allowed values. If you are strictly against that, please let's discuss, because I don't think I understand the rationale.

internal/release/promote/pr_rendering.go
silphid: Nice, TIL about `strings.Cut()`! :)

internal/release/promote/pr_rendering.go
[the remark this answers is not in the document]
silphid: TIL!

davidmdm (approved):

internal/release/promote/pr_rendering.go
ayahajar: for swapping the 2 values, I would still use the shortcut 
  sourceTag, targetTag = targetTag, sourceTag

PR #113, opened 2024-03-08
refactor(PL-2532): better-git-errors

silphid:
Many of our git errors are esoteric and unactionable, such as:
```
Error: getting default branch: exit status 128
```
This PR includes the combined output of the failing command in the error message, so that we can better troubleshoot issues.


riclima (approved):

alexstojda (changes requested):
This should be a `fix` not `refactor` as the latter will not cause a new version to be released.

silphid:
@alexstojda @riclima see new PR: https://github.com/nestoca/joy/pull/114
(didn't know how to rename branch with same PR)

PR #118, opened 2024-03-11
fix(PL-2541): fetch all labels first

j-martin:


davidmdm (approved):

j-martin:
Yeah, the issue was that the default limit is 30 when fetching labels and Joy has been creating a lot of labels in the past few days, so we started fetching only a subset of labels. Basically, we have a label per release (aka project aka service) and environment.

We need the full list to decide to create the label or not: https://github.com/nestoca/joy/blob/49095fe2abc965527a8348d93a44e9a6f458ec06/internal/github/pull_request_provider.go#L217-L220

We could also add `--force` to the label creation if we ever have more than 1000 labels. 😬

PR #121, opened 2024-03-12
feat(PL-2418): add !local tag annotation

davidmdm:
This PR adds `!local` annotations for values that should be locked, but should not have their structure promoted across environment boundaries.

### DIFF
--- a/internal/yml/merge.go
+++ b/internal/yml/merge.go
@@ -34,7 +34,7 @@ func Merge(dst, src *yaml.Node) *yaml.Node {
 func merge(dst, src *yaml.Node) *yaml.Node {
 	// If destination is locked, it does not matter what source is.
 	// If destination exists but source is locked, disregard source.
-	if isLocked(dst) || (dst != nil && isLocked(src)) {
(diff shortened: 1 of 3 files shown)


j-martin (approved):
I don't mind the name `local` (not "perfect", but better than my suggestions), but we MUST document this in the README.md and Confluence if applicable.

Pre-approving pending docs.

silphid (approved):
Great job @davidmdm !  I would recommend adding a second merge on top of first merge (actual=src>dest, actual=src>actual) in your test `t.Run()` block to ensure that all tests cases are idempotent. Eg:
```
actual, err := yaml.Marshal(Merge(&dst, &src))
require.NoError(t, err)
actual, err = yaml.Marshal(Merge(&actual, &src))
```
but maybe it would be worth check `actual` against `expected` for each step individually, so that we catch normal issues earlier and independently from non-idempotent issues (could even mention "non-idempotency" explicitly in assert failure message).

davidmdm:
@silphid Added an idempotency check to the test cases! good idea!

PR #122, opened 2024-03-13
feat(PL-2549): make version compatible with go toolchain

davidmdm:


riclima (approved):

cmd/joy/main.go
```diff
@@ -2,14 +2,24 @@ package main
...
+	}
+
 	if err := NewRootCmd(version).Execute(); err != nil {
 		os.Exit(1)
 	}
 }
+
+func debugBuildVersion() string {
+	info, _ := debug.ReadBuildInfo()
+	return info.Main.Version
```
j-martin, line 24: So this would return `latest` or `(devel)` if we used `go install github.com/nestoca/joy@latest`? If that's the later that is fine/good.

If we use `go install github.com/nestoca/joy@v0.34.5` it returns `v0.34.5`? Is so that's great.

j-martin (approved):
Approving assuming my assumption(s) below is correct.

cmd/joy/main.go
davidmdm: It will do that for all builds after this new release.

The Gotoolchain embes the build info into the binary. And when it builds from a remote source it uses the tag as a version. However our old binaries don't use this information but instead read from the variable that we overwrite with LDFLAGS.

But our new binaries will use the debug info when used wth the go toolchain

PR #124, opened 2024-03-13
feat(PL-2470): add joy release links command

silphid:


j-martin (commented):
I've played around with it. It seems to be working well enough!

I think we should be able to [add links to projects and releases](https://github.com/nestoca/catalog/pull/1448#discussion_r1524230297)

Side note (as in, in another PR/ticket), I think we should improve how the fuzzy finder works.

Right now the user is expected to type the exact and correct arguments, but it would be nice if it worked like `fzf` (example below).

```
go run ./cmd/joy release open --env prod acco actio
Error: environment "prod" not found
exit status 1
```

It would already be much more useful if the potential choices would be outputted.

```
$ echo '123\nabc\ndef' | fzf --select-1 --query b
abc
```

davidmdm (approved):

PR #131, opened 2024-03-22
feat/PL 2554/support chart refs

davidmdm:
This PR adds support for chart refs:

# Changes

## Features

Going forward known charts will be configured from the catalog's `joy.yaml` and be referenced by environments and releases.

Releases will retain the ability to define their own chart but if they do so must define the chart in full.
If using a chart reference, release's may override the desired version.
Environments can control the default version of a ref for that environment.

## Refactoring

This PR touches a lot of code because we've moved the catalog to be loaded at startup within the prerun of the root command, it is then passed via context to the subcommands. This means that there's only one point entry for catalog loading, and that the catalog no longer has load options, simplifying the use of the catalog and allowing for tests to have tests at the command level without a test filesystem.

This refactoring was necessary since the change to support chart refs needed to modify how the catalog was loaded in every command, and this was too tedious.


silphid (approved):
LGTM! 👍

alexstojda:
@davidmdm you have a non-conventional commit in here, please be sure to edit the commit message, or squash merge :) 

(Still reviewing)

alexstojda (changes requested):
Just a small nit about modifying the CHANGELOG

Also, it would be great if you could include a summary / description of the changes in the PR, especially when the changes are this big :)

PR #134, opened 2024-03-25
fix(PL-2582): fix build promote

j-martin:
1d3c9fb4350cc5a18e33bc78e15ea59e02da57d5 broke `build promote`    in a few ways:

 - Did not have a writer

 - Releases items now are in an array where the index is set per environment. Not all environments may have a release. Note: Why not using a map?

### DIFF
--- a/cmd/joy/build.go
+++ b/cmd/joy/build.go
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/nestoca/joy/internal/build"
+	"github.com/nestoca/joy/internal/yml"
 	"github.com/nestoca/joy/pkg/catalog"
 )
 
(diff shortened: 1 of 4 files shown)


davidmdm (approved):
The first point is correct. This was a change that wasn't caught. We need E2E tests for build promote.

The latter has always been like this. So I would test that this works with the applications release because I don't know why it would have worked before. 

If it works, then let's merge!

j-martin:
So in the old version of the code, it would only return one item (for the environment). With your change it became one per env with release.

Old (before https://github.com/nestoca/joy/commit/1d3c9fb4350cc5a18e33bc78e15ea59e02da57d5)

image

New (https://github.com/nestoca/joy/commit/1d3c9fb4350cc5a18e33bc78e15ea59e02da57d5)

image

PR #137, opened 2024-03-26
fix/PL 2584/fix rel prom

davidmdm:
- **fix(PL-2584): fix withEnvironments and build promote**
- **fix(PL-2584): fix withEnvironments and build promote**
- **fix(PL-2584): fix rel prom**


silphid (approved):
LGTM! 👍

PR #140, opened 2024-03-27
fix(PL-2557): use moq for chart rendering tests

davidmdm:


alexstojda (approved):
This was supposed to be for next week ;P

PR #141, opened 2024-03-27
feat(PL-2530): add joy extensions

silphid:
See documentation:
* [How to list and execute joy extension scripts](https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/2669936744/Internal+Developer+Platform#How-to-list-and-execute-joy-extension-scripts)
* [How to create your own joy extension scripts](https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/2669936744/Internal+Developer+Platform#How-to-create-your-own-joy-extension-scripts)

### DIFF
--- a/cmd/joy/execute.go
+++ b/cmd/joy/execute.go
@@ -1,0 +1,186 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/nestoca/survey/v2"
+
+	"github.com/nestoca/joy/internal/config"
+
+	"github.com/spf13/cobra"
+)
+
+func NewExecuteCmd() *cobra.Command {
+	var list bool
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:     "execute",
+		Aliases: []string{"exec", "x"},
+		Args:    cobra.ArbitraryArgs,
+		Short:   "Execute user joy scripts",
+		Long:    `Execute custom user joy scripts.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg := config.FromContext(cmd.Context())
+			searchPath := os.Getenv("PATH")
+
+			scripts, err := getScripts(cfg.CatalogDir, searchPath)
+			if err != nil {
+				return fmt.Errorf("getting scripts: %w", err)
+			}
+
+			if list {
+				if jsonOutput {
+					output, err := formatScriptsAsJson(scripts)
+					if err != nil {
+						return fmt.Errorf("formatting scripts as JSON: %w", err)
+					}
+					_, err = fmt.Fprintln(cmd.OutOrStdout(), output)
+					return err
+				}
+
+				_, err = fmt.Fprintln(cmd.OutOrStdout(), formatScripts(scripts))
+				return err
+			}
+
+			script, err := func() (Script, error) {
+				if len(args) == 0 {
+					script, err := selectScript(scripts)
+					if err != nil {
+						return Script{}, fmt.Errorf("selecting script: %w", err)
+					}
+					return script, nil
+				}
+
+				script, err := getScript(scripts, args[0])
+				if err != nil {
+					return Script{}, fmt.Errorf("getting script: %w", err)
+				}
+				args = args[1:]
+				return script, nil
+			}()
+
+			return execute(cfg.CatalogDir, script.Path, args)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&list, "list", "l", false, "List available scripts")
+	cmd.Flags().BoolVarP(&jsonOutput, "json", "j", false, "Output as JSON")
+
+	return cmd
+}
+
+func selectScript(scripts []Script) (Script, error) {
+	var scriptNames []string
+	for _, script := range scripts {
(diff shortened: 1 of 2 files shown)


cmd/joy/execute.go
```diff
@@ -0,0 +1,184 @@
...
+				return fmt.Errorf("getting scripts: %w", err)
+			}
+
+			if list {
+				if jsonOutput {
+					output, err := formatScriptsAsJson(scripts)
+					if err != nil {
+						return fmt.Errorf("formatting scripts as JSON: %w", err)
+					}
+					fmt.Println(output)
```
davidmdm, line 43: fmt.Fprintln(cmd.OutOrStdout(), ...)

cmd/joy/execute.go
```diff
@@ -0,0 +1,184 @@
...
+					return nil
+				}
+
+				fmt.Println(formatScripts(scripts))
+				return nil
+			}
+
+			var script Script
```
davidmdm, line 51: looks like a beautiful place for an IIFE

cmd/joy/execute.go
```diff
@@ -0,0 +1,184 @@
...
+			continue
+		}
+		name := file.Name()
+		scriptsMap[name] = Script{
+			Name: name,
+			Path: filepath.Join(catalogBinDir, file.Name()),
+		}
+	}
+
+	searchDirs := filepath.SplitList(searchPath)
```
davidmdm, line 115: I would recommend KISS: joy is wired to work with a catalog. Joy extensions should only live in the catalog, as the extensions are meant to extend the behaviour of that catalog.  Let's not look for scripts in the user's PATH.

davidmdm: In that sense we don't even need "joy" prefixes.

davidmdm: Then we don't even need a type Script, it would just be the filenames under `catalog/bin`

cmd/joy/execute.go
```diff
@@ -0,0 +1,184 @@
...
+			return script, nil
+		}
+	}
+	return Script{}, fmt.Errorf("script %q not found", scriptName)
+}
+
+func execute(catalogDir, scriptPath string, args []string) error {
+	cmd := exec.Command(scriptPath, args...)
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, fmt.Sprintf("JOY_CATALOG_DIR=%s", catalogDir))
```
davidmdm, line 179:
```
cmd.Env = append(os.Environ(), ...)
```

cmd/joy/execute.go
silphid: I disagree, part of the goal is to allow power users to create their own scripts! 🤷

silphid: And that loosely follows the same way that `kubectl` extensions can be defined, by defining them with the `kubectl-` prefix and placing them in your `PATH`.

davidmdm (approved):

cmd/joy/execute.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+func NewExecuteCmd() *cobra.Command {
+	var list bool
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:     "execute",
+		Aliases: []string{"exec", "x"},
```
j-martin, line 24: `e` is not used, is your intent to use it for `environment`?

```suggestion
		Aliases: []string{"exec", "e"},
```

cmd/joy/execute.go
davidmdm: I really don't see the difference for a power user between invoking an extension like so:
```
joy exec extension
---
joy-extension
```

I am not sure the environment variable injection adds much value, and it basically encourages users to have private scripts instead of developing useful utilities in the public eye.

Less is more. More is only ever more, when it is less.

j-martin: You better check my `PATH`! ;) 

`git` also did the same thing well before k8s was a thing.

cmd/joy/execute.go
```diff
@@ -0,0 +1,186 @@
...
+		if script.Name == scriptName {
+			return script, nil
+		}
+	}
+	return Script{}, fmt.Errorf("script %q not found", scriptName)
+}
+
+func execute(catalogDir, scriptPath string, args []string) error {
+	cmd := exec.Command(scriptPath, args...)
+	cmd.Env = append(os.Environ(), fmt.Sprintf("JOY_CATALOG_DIR=%s", catalogDir))
```
j-martin, line 181: That seems to be a good enough way to tell the script where we are running. I was debating if the working directory should be the catalog dir, but your approach is more flexible.

I'll update `k8s-apply-vpa-recommendations.sh` once this is merged.

cmd/joy/execute.go
silphid: Indeed. Though it unifies user scripts with catalog scripts, makes them discoverable with `joy exec --list`, and automatically injects the `JOY_CATALOG_DIR` env var.  And eventually we can pass more context to those scripts.

cmd/joy/execute.go
silphid: Yeah, `e` will be used for `environments`, for one-letter command lovers like yourself! 😄

PR #143, opened 2024-03-27
feat(PL-2602): joy rel ls --json

silphid:
See documentation [here](https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/2669936744/Internal+Developer+Platform#How-to-create-your-own-joy-extension-scripts) for an example script that uses `joy rel ls --json`.

Actually, here it is:
```
#!/usr/bin/env bash
joy release list --env staging --json 2>/dev/null | 
    jq -r '.crossReleases[] | select(.name == "accounts") | .releases[] | select(.status == "ahead") | "\(.metadata.name): \(.version)"'
```

It would be great to internally do the equivalent of `2>/dev/null` for `--json` output, otherwise it clutters up the output (even if it's now sent to `stderr`, the user still sees it).

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -60,9 +60,11 @@ func NewReleaseCmd() *cobra.Command {
 func NewReleaseListCmd() *cobra.Command {
 	var releases, envs, owners string
 	var narrow, wide bool
+	var jsonOutput bool
 	cmd := &cobra.Command{
 		Use:     "list",
-		Aliases: []string{"ls"},
+		Aliases: []string{"ls", "l"},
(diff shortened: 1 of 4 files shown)


j-martin (approved):
I didn't ask for it, but I like it!

davidmdm (approved):
Although I stand by my comment, if it works it works.

silphid:
> I think building all this information in the return of the list is not necessary
Again, let me disagree. That's precisely the kind of information we need to create good scripts and automations, like my example in the description. Otherwise, the json output is not that useful really.

PR #146, opened 2024-04-02
test(PL-2576): add joy build promote e2e tests

silphid:


internal/build/promote.go
```diff
@@ -51,7 +51,6 @@ func Promote(opts Opts) error {
 		promotionCount++
 	}
 
-	// Print summary
```
davidmdm, line 54: ❤️

davidmdm (approved):
The test cases are good.

My only nit is that for error condition tests (the ones that do not write to disk but will return an error), it would be more elegant to code the catalog in memory instead of from an FS.

Approved nonetheless!

PR #147, opened 2024-04-02
fix(PL2614): update catalog and config on catalog pull

davidmdm:
This PR updates the EnsureCleanAndUpToDateWorkingCopy function such that it will update the config and catalog context objects.

The one issue with this approach is that we don't run the same checks as the root, such as checking joy min Versions, etc.

However we weren't doing so before, so this is not a regression. But we will need to think on it for a future task.

### DIFF
--- a/internal/git/clean_working_copy.go
+++ b/internal/git/clean_working_copy.go
@@ -1,24 +1,15 @@
 package git
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"strings"
 
-	"github.com/nestoca/joy/internal/config"
 	"github.com/nestoca/joy/internal/style"
 )
 
-func EnsureCleanAndUpToDateWorkingCopy(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


silphid (approved):

internal/git/clean_working_copy.go
```diff
@@ -1,24 +1,15 @@
...
 	"os"
 	"strings"
 
-	"github.com/nestoca/joy/internal/config"
 	"github.com/nestoca/joy/internal/style"
 )
 
-func EnsureCleanAndUpToDateWorkingCopy(ctx context.Context) error {
-	if config.FlagsFromContext(ctx).SkipCatalogUpdate {
-		_, _ = fmt.Fprintln(os.Stderr, "ℹ️ Skipping catalog update and dirty check.")
```
silphid, line 15: J'aurais peut-être juste gardé ce print info?

silphid (approved):
Génial! Juste un commentaire mineur...

PR #151, opened 2024-04-03
feat(PL-2615): add support for bulk promotions

davidmdm:


cmd/joy/release_promote_filters_test.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+
+			for _, cross := range cat.Releases.Items {
+				for _, rel := range cross.Releases {
+					if rel == nil {
+						continue
+					}
+					require.NoError(t, yaml.Unmarshal(rel.File.Yaml, &rel))
+				}
+			}
```
silphid, line 240: Wouldn't it be possible to merge those two loops?

cmd/joy/release_promote_filters_test.go
```diff
@@ -0,0 +1,222 @@
...
+				{
+					Name:   "pre-release-target",
+					Source: &TestFile{Path: "alpha-source.yaml", Content: "{spec: {version: 1.2.3}}"},
+					Target: &TestFile{Path: "alpha-target.yaml", Content: "{spec: {version: 1.0.0-some-feature}}"},
+				},
+				{
+					Name:   "nil-target",
+					Source: &TestFile{Path: "beta-source.yaml", Content: "{spec: {version: 3.2.1}}"},
+					Target: nil,
+				},
```
silphid, line 158: Maybe just add a comment to explain why this nil target matters in this test? Otherwise, it may look like testing two different cases in the same test, whereas it's really related. Or maybe it could be broken down into two separate tests to make things clearer(?)

silphid (approved):
Great work! 👍  Just a few comments, but still approving!

cmd/joy/release.go
```diff
@@ -222,9 +241,12 @@ func NewReleasePromoteCmd(preRunConfigs PreRunConfigs) *cobra.Command {
 	cmd.Flags().StringSliceVar(&templateVars, "template-var", nil, "Variable to pass to PR/commit templates, in the form of key=value (flag can be specified multiple times)")
 	cmd.Flags().BoolVarP(&narrow, "narrow", "n", false, "Use narrow columns mode")
 	cmd.Flags().BoolVarP(&wide, "wide", "w", false, "Use wide columns mode")
+	cmd.Flags().BoolVar(&all, "all", false, "select all releases")
+	cmd.Flags().BoolVar(&keepPreRelease, "keep-pre-release", false, "do not promote releases that are pre-releases in target env")
```
j-martin, line 245: With the usage of `--omit` should we call this `--omit-pre-release`?

```suggestion
	cmd.Flags().BoolVar(&keepPreRelease, "omit-pre-release", false, "Do not promote releases that are pre-releases in target env")
```

cmd/joy/release_promote_filters_test.go
davidmdm: It is totally possible but it hurts readability in my opinion

cmd/joy/release.go
```diff
@@ -222,9 +241,12 @@ func NewReleasePromoteCmd(preRunConfigs PreRunConfigs) *cobra.Command {
...
+	cmd.Flags().StringSliceVar(&omit, "omit", nil, "releases to omit from promotion")
```
j-martin, line 246: Casing

```suggestion
	cmd.Flags().StringSliceVar(&omit, "omit", nil, "Releases to omit from promotion")
```

cmd/joy/release.go
```diff
@@ -222,9 +241,12 @@ func NewReleasePromoteCmd(preRunConfigs PreRunConfigs) *cobra.Command {
 	cmd.Flags().StringSliceVar(&templateVars, "template-var", nil, "Variable to pass to PR/commit templates, in the form of key=value (flag can be specified multiple times)")
 	cmd.Flags().BoolVarP(&narrow, "narrow", "n", false, "Use narrow columns mode")
 	cmd.Flags().BoolVarP(&wide, "wide", "w", false, "Use wide columns mode")
+	cmd.Flags().BoolVar(&all, "all", false, "select all releases")
```
j-martin, line 244: I wonder if it would make more sense to pass this as a positional arg like a release, where if `release=all`, then load all.

cmd/joy/release.go
davidmdm: omit is not about prereleases, its about omitting any release. So you could do:

```
joy rel prom --all --omit servicing
```

cmd/joy/release.go
davidmdm: I know some apis do this, but I don't love it. Now your positional args has a special name that releases can't have. 
I know its not a name we would ever use. But still. The dashes to make it a flag is tried and true!

cmd/joy/release_promote_filters_test.go
silphid: If there's anything, current implementation left me more confused, not understanding why looping twice over the same items.  Here's what it would look like, not finding it less readable really:
```go
			for _, release := range tc.Releases {
				crossRel := cross.Release{
					Name: release.Name,
					Releases: []*v1alpha1.Release{
						v1alpha1Release(release, source, release.Source),
						v1alpha1Release(release, target, release.Target),
					},
				}
				cat.Releases.Items = append(cat.Releases.Items, &crossRel)
				for _, rel := range crossRel.Releases {
					if rel == nil {
						continue
					}
					require.NoError(t, yaml.Unmarshal(rel.File.Yaml, &rel))
				}
			}
```

cmd/joy/release.go
davidmdm: I find the directionality very confusing.

Are we omitting prereleases from the source? Are omitting updates to the target if its a prerelease?

To me "keep" or "preserve" indicates to my brain that we are talking about the target environment... Although the english language is ambiguous.

silphid: I'd be OK with that, sorry David I know it's my fault we went with "pre-release"! ;)

cmd/joy/release.go
silphid: Not super in favour of special-meaning names, we successfully were able to avoid those up to now! 😁 `--all` also has the great advantage of being self-documenting, as opposed to special names.

cmd/joy/release.go
j-martin: > Are we omitting prereleases from the source? Are omitting updates to the target if its a prerelease?

You got a point, I would argue that is why there is a flag description. 

In practice, QA (the source in 99% of cases) has only regular releases. 🤷‍♂️

silphid: I admit that "keep" has no ambiguity, whereas "omit" definitely has

cmd/joy/release_promote_filters_test.go
davidmdm: I guess it is a matter of opinion. I will change it!

cmd/joy/release.go
```diff
@@ -126,23 +127,34 @@ func NewReleaseListCmd(preRunConfigs PreRunConfigs) *cobra.Command {
...
-		Args:    cobra.RangeArgs(0, 1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if autoMerge && draft {
 				return fmt.Errorf("flags --auto-merge and --draft cannot be used together")
 			}
 			if noPrompt {
-				if len(args) == 0 {
-					return fmt.Errorf("releases are required when no-prompt is set")
+				if len(args) == 0 && !all {
+					return fmt.Errorf("one of releases or all are required when no-prompt is set")
```
j-martin, line 157:
```suggestion
					return fmt.Errorf("one of releases or --all are required when no-prompt is set")
```

cmd/joy/release.go
j-martin: My thought it that both the `--keep-prerelease` and `--omit` both more or less do the same thing (preventing promotions), but have different names:

`... --all --keep-prerelease --omit accounts --omit salesforce`

That being said, if you guys are adamant about it, I won't block you.

cmd/joy/release.go
j-martin: Fair enough, we got a check to only accept the positional arg(s) or the flag, so we should be good.

https://github.com/nestoca/joy/pull/151/files?w=1#diff-decfa809d626e940d8f9f17e4be211eb2af9b99e02cbd12d93500b3e5cc73a3eR156

PR #157, opened 2024-04-09
test(PL-2576): add joy release promote e2e tests

silphid:


cmd/joy/release_promote_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+func TestPromoteAllReleasesFromStagingToProd(t *testing.T) {
+	dir, output, err := executeReleasePromoteCommand(t, NewReleasePromoteCmd(PromoteParams{PreRunConfigs: make(PreRunConfigs)}),
+		"--source", "staging",
+		"--target", "prod",
+		"--no-prompt",
+		"--draft",
+		"--template-var", "triggeredBy=john.doe",
+		"foo")
+	fmt.Println(output)
```
davidmdm, line 57: console.log lifestlye!

cmd/joy/release_promote_test.go
```diff
@@ -0,0 +1,205 @@
...
+	require.NoError(t, err)
+
+	prURLRegex := regexp.MustCompile(`Created draft pull request: https://github.com/nestoca/joy-release-promote-test/pull/([0-9]+)`)
+	match := prURLRegex.FindStringSubmatch(output)
+	require.NotNil(t, match)
+	prURL := match[1]
+	t.Cleanup(func() {
+		closePR(t, dir, prURL)
```
davidmdm, line 65: I think in this case, you don't need to capture the variables in the closure. 

This should do the trick:
```go
defer closePR(t, dir, prURL)
```

cmd/joy/release_promote_test.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+func executeReleasePromoteCommand(t *testing.T, cmd *cobra.Command, args ...string) (string, string, error) {
+	dir := testutils.CloneToTempDir(t, "joy-release-promote-test")
+
+	cat, err := catalog.Load(dir, nil)
+	assert.NoError(t, err)
+
+	cfg, err := config.Load(dir, dir)
+	assert.NoError(t, err)
```
ayahajar, line 36: Let's stick to using require only

davidmdm: Big plus one to this!

cmd/joy/release_promote_test.go
silphid: What's the difference between `t.Cleanup()` and `defer`, any clue why they introduced `t.Cleanup()`?

davidmdm (approved):

cmd/joy/release_promote_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	branchRegex := regexp.MustCompile(`Committed and pushed new branch (promote-foo-from-staging-to-prod-([0-9a-f-]+)) with message:`)
+	match = branchRegex.FindStringSubmatch(output)
+	require.NotNil(t, match, "Branch name should be specified in output:\n"+output+"\n---")
+	branch := match[1]
+
+	require.NoError(t, git.Checkout(dir, branch))
+	actual, err := os.ReadFile(filepath.Join(dir, "environments/prod/foo.yaml"))
+	require.NoError(t, err)
+	expected := `apiVersion: joy.nesto.ca/v1alpha1
```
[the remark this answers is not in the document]
silphid: Using string comparison here, so we would need to parse both actual and expected strings to perform comparison if that were JSON...

cmd/joy/release_promote_test.go
davidmdm: So I had to read about this. `t.Cleanup` is needed for tests marked with t.Parrallel(), because the testing framework needs to be aware of the subtests. 

The function passes through all of the subtests and registers them to run, if all subtests are marked as parrallel they will continue async and the main function will exit before the subtests finish and so parent defers would run before their subtests finish potentially.

Hence, t.Cleanup is to be used for scheduling cleanup when you have parrallel subtests.

It's quite `niché`

cmd/joy/release_promote_test.go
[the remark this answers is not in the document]
davidmdm: This is going to my list of why we cannot have nice things.

Require package has `require.EqualJSON` which allows you to compare two strings as equivalent json objects, we would need to write the equivalent `EqualYAML` helper, which is probably not justified just to please my eyes.

ayahajar (approved):

PR #159, opened 2024-04-10
test(PL-2576): add joy pr promote e2e tests

silphid:
* Added e2e test for `joy pr promote`
* Introduced a `CatalogBuilder` to help us avoid defining test catalogs in-memory in a concise yet flexible way.
* Added `--no-prompt`, `--target-env` and `--disable` flags to `joy pr promote` command to make it testable, but also usable non-interactively.
* Made `joy pr promote` command testable, with parametrizable stdout
* Refactored tests for links to use `CatalogBuilder`


api/v1alpha1/environment.go
```diff
@@ -136,7 +136,7 @@ func GetEnvironmentsByNames(environments []*Environment, names []string) []*Envi
 	return envs
 }
 
-func (e *Environment) IsPromotableTo(targetEnv *Environment) bool {
+func (e Environment) IsPromotableTo(targetEnv *Environment) bool {
```
davidmdm, line 139: If we are unifying the receiver names, can we agree to prefer short but non letter names? I think `env.Spec` reads much better than `e.Spec`.

The only exception I believe in is the actual conventions like `w` for io.Writer, `r` for io.Reader, b for `bytes.Buffer` etc.

ayahajar: I personally go for letters with receivers but I'm fine with both

silphid: I'm OK with that, but we already have a huge amount of 1-letter receivers! 🤷

davidmdm (approved):

PR #161, opened 2024-04-19
feat(PL-2637): custom error help messages

silphid:
Use this dummy catalog draft PR to test with:
https://github.com/nestoca/catalog/pull/2528

* Try `joy --help` or any other command help to display help with generic message added at the end:
```
See: https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/2669936744/Internal+Developer+Platform#How-to-cross-promote-a-release-from-qa-to-production
```

* Try `joy rel prom --help` to display message for that specific command:
```
See: https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/2669936744/Internal+Developer+Platform#How-to-cross-promote-a-release-from-qa-to-production
```

* Try `joy env open dummy` to trigger an `environment "dummy" not found` error and display the following message specifically intended for that error:
```
The "dummy" environment has been removed, please use "gimmick" instead.
```


internal/help/help.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/nestoca/joy/internal/config"
+)
+
+const allCommandsKey = "_all"
```
davidmdm, line 13: This kills me to have to nit this again. But can the prefix be any symbol that's not considered a word character? Word characters are `a-zA-Z_` also I would consider a `-` to not be a good candidate for this. Probably same for `$` but less so.  

`!` | `@` | `#` | `&` | `*` | `~` would all be fantastic choices.

cmd/joy/main.go
```diff
@@ -1,22 +1,56 @@
...
+	cfg, err := config.Load(configDir, catalogDir)
+	if err != nil {
+		return err
+	}
+	rootCmd.SetContext(config.ToContext(context.Background(), cfg))
+	help.AugmentCommandHelp(rootCmd)
+
+	cmd, err := rootCmd.ExecuteC()
+	if err != nil {
+		err = help.WrapError(cmd, err)
```
davidmdm, line 50: nit: 
```go
return help.WrapError
```

internal/help/help.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func AugmentCommandHelp(cmd *cobra.Command) {
+	baseHelpFunc := cmd.HelpFunc()
+	cmd.SetHelpFunc(func(cmd *cobra.Command, args []string) {
+		baseHelpFunc(cmd, args)
+
+		message, err := getMessageFromCommand(cmd, nil)
+		if err != nil {
+			panic(err)
```
davidmdm, line 38: I probably wouldn't panic here. Would be a weird experience to run a command with `--help` and have it panic.

Might as well write it to stderr.

internal/help/help.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		if message != "" {
+			fmt.Println("\n" + message)
```
davidmdm, line 41: I suspect the message should go to stderr! Pretty sure help text goes there by default, and so the text is now interleaved between two file descriptors. Though correctly formatted in the terminal for sure :)

internal/help/help.go
```diff
@@ -0,0 +1,85 @@
...
+				continue
+			}
+		}
+		messages = append(messages, strings.TrimSpace(help.Message))
+	}
+
+	return strings.Join(messages, "\n\n"), nil
+}
+
+func getFullCommandName(cmd *cobra.Command) string {
```
davidmdm, line 77: From memory, I think this is already implemented by cobra. 
cobra.PathCommand or something like that. Might need to strip away the `joy ` prefix but thats trivial.

```go
strings.TrimPrefix(cmd.PathCommand(), "joy ")
```

internal/help/help.go
davidmdm: EDIT: `&` and `*` would be poor choices since those are anchor tags in yaml documents

davidmdm (approved):

PR #166, opened 2024-04-25
chore(PL-2662): separate user and catalog configurations

davidmdm:


internal/environment/configure_selection.go
```diff
@@ -11,18 +12,18 @@ import (
 
 func ConfigureSelection(cat *catalog.Catalog, configFilePath string, all bool) error {
 	// Load fresh copy of config file, without any alterations/defaults applied
-	cfg, err := config.LoadFile(configFilePath)
+	cfg, err := config.Load(filepath.Dir(configFilePath), "")
```
silphid, line 15: Just want to make sure this load func does not merge in anything else than the bare config file itself, as we don't want to persist back anything that was not in the originally loaded file. Have you tested it manually just to sanity check? That's the most important thing that popped to my mind.

silphid (approved):
LGTM! Just make sure to check my comments.

internal/environment/configure_selection.go
davidmdm: This loads a fresh copy of the entire config object, and modifies and saves the user config back separately from everything else, I believe it to be safe!

internal/config/config.go
```diff
@@ -57,23 +63,23 @@ type Config struct {
...
 	Templates Templates `yaml:"templates,omitempty"`
 
-	ColumnWidths ColumnWidths `yaml:"columnWidths,omitempty"`
-
 	Helps map[string][]Help `yaml:"help,omitempty"`
 }
 
+type Config struct {
+	User    `yaml:",inline"`
+	Catalog `yaml:",inline"`
```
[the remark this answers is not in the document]
davidmdm: Nice catch, I don't ever deserialize directly to this struct.

internal/environment/configure_selection.go
davidmdm: I'll make it nicer though

PR #168, opened 2024-05-01
fix(PL-2680): exclude gha bot from reviewers

silphid:


internal/github/pull_request_provider.go
```diff
@@ -81,6 +81,9 @@ func (p *PullRequestProvider) Create(params pr.CreateParams) (string, error) {
 	}
 
 	for _, reviewer := range params.Reviewers {
+		if reviewer == "github-actions[bot]" {
+			continue
+		}
```
alexstojda, line 86: Should this be hardcoded? Maybe some other workflows will use another bot name.

alexstojda (approved):
Once comment, else LGTM

internal/github/pull_request_provider.go
silphid: Yeah, that's possible, but we're assuming that's the only one that GHA uses. We went for the simplest solution to address the issue promptly. If we realize there is a wealth of different GH users we may want to exclude, we'll introduce a catalog-level config to specify them. But let's see first, if you're okay with that.

PR #170, opened 2024-05-01
feat(PL-2679): add support for environment-level values

silphid:


internal/release/render/render.go
```diff
@@ -197,6 +203,110 @@ func HydrateValues(release *v1alpha1.Release, mappings *config.ValueMapping) (ma
...
+	case string:
+		operator, resolvedValue, err := resolveOperatorAndValue(val, envValues)
+		if err != nil {
+			return nil, err
+		}
+		if operator != "" && operator != "ref" {
+			return nil, fmt.Errorf("only $ref() operator supported within object: %s", val)
+		}
+		return resolvedValue, nil
+	case map[string]any:
```
davidmdm, line 229: probably need to duplicate the logic of `map[string]any` for `map[any]any` because the value is deserialized from yaml not json.

davidmdm (approved):

davidmdm (changes requested):
Just realized we need to make sure to hydrate values before doing validation. This will need to be added to PR

api/v1alpha1/environment.go
```diff
@@ -54,6 +54,10 @@ type EnvironmentSpec struct {
 	// SealedSecretsCert is the public certificate of the Sealed Secrets controller for this environment
 	// that can be used to encrypt secrets targeted to this environment using the `joy secret seal` command.
 	SealedSecretsCert string `yaml:"sealedSecretsCert,omitempty" json:"sealedSecretsCert,omitempty"`
+
+	// Values are the environment-level values that can optionally be injected into releases' values during rendering
+	// via the `$ref(.Environment.Spec.Values.someKey)` or `$spread(...)` template expressions.
+	Values map[string]interface{} `yaml:"values,omitempty" json:"values,omitempty"`
```
davidmdm, line 60: I can feel the copilot generation that used `interface{}` instead of `any`

internal/release/render/render.go
```diff
@@ -7,13 +7,15 @@ import (
...
+	"regexp"
 	"slices"
 	"strings"
 
+	"github.com/Masterminds/sprig/v3"
+
 	"github.com/TwiN/go-color"
 	"github.com/nestoca/survey/v2"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
```
davidmdm, line 18: found it!

davidmdm (approved):
Approving but please fix yaml import before merging!

PR #171, opened 2024-05-08
feat(PL-2701): unify cue schema with values

davidmdm:
This PR changes Joy's value hydration strategy to also unify against the chart's `values.cue`. In this way the defaults that we provide in nested structures of our values.cue file will be applied to the values we provide.


silphid:
Nice! What's left?

internal/release/render/render.go
```diff
@@ -200,6 +205,44 @@ func HydrateValues(release *v1alpha1.Release, mappings *config.ValueMapping) (ma
...
+	runtime := cuecontext.New()
+
+	schema := runtime.
+		CompileBytes(rawSchema).
+		LookupPath(cue.MakePath(cue.Def("#values")))
+
+	unified := schema.Unify(runtime.Encode(values))
+
+	if err := unified.Validate(cue.Final(), cue.Concrete(true)); err != nil {
+		return nil, xerr.MultiErrFrom("validating values", AsErrorList(cueerrors.Errors(err))...)
```
silphid, line 238: Shouldn't that be "unifying" instead of "validating"?
```suggestion
		return nil, xerr.MultiErrFrom("unifying values", AsErrorList(cueerrors.Errors(err))...)
```

internal/release/render/render.go
```diff
@@ -200,6 +205,44 @@ func HydrateValues(release *v1alpha1.Release, mappings *config.ValueMapping) (ma
...
+	}
+
+	var result map[string]any
+	if err := unified.Decode(&result); err != nil {
+		return nil, fmt.Errorf("decoding value: %w", err)
```
silphid, line 243:
```suggestion
		return nil, fmt.Errorf("decoding values: %w", err)
```

internal/release/validate/validate.go
```diff
@@ -91,48 +81,3 @@ func ValidateRelease(ctx context.Context, params ValidateReleaseParams) error {
 
 	return nil
 }
-
-func validateSchema(release *v1alpha1.Release, mappings *config.ValueMapping, chart *helm.ChartFS) error {
```
silphid, line 95: I'm assuming this is going away because validation is now implicitly achieved via unification, right?

silphid (approved):

internal/release/render/render.go
davidmdm: This is a validation error!

internal/release/validate/validate.go
davidmdm: Yes, we run a concrete validation before decoding

internal/release/render/render.go
silphid: Effectivement, désolé, j'avais mal regardé! 😅

j-martin:
General comment: There is no description in the ticket, the PR, or the commit of what this change does. Can we write something down somewhere?

silphid (approved):

PR #175, opened 2024-05-21
feat(PL-2723): validate locked todos

davidmdm:


internal/yml/todos_test.go
```diff
@@ -0,0 +1,62 @@
...
+		Expected bool
+	}{
+		{
+			Name: "no locked todos",
+			Value: `{
+				hello: TODO, # not locked
+				array: [1, 2, foo, todo, TODO], # todos are still not locked!
+				map: {
+					value: TODO,
+					john: doe,
```
j-martin, line 25: For exhaustiveness' sake, we could separate those as different test cases.

j-martin (approved):
I would split the test case, but LGTM otherwise.

PR #177, opened 2024-05-22
feat(PL-2729): Support ignoring files when loading catalog

alexstojda:
- **chore(PL-2729): Import ignorefile parsing library**
- **refactor(PL-2729): Adjust some naming in the imported package**
- **feat(PL-2729): Support ignoring files when loading catalog**

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -14,6 +14,7 @@ import (
 	"gopkg.in/godo.v2/glob"
 
 	"github.com/nestoca/joy/api/v1alpha1"
+	"github.com/nestoca/joy/internal/ignore"
 	"github.com/nestoca/joy/internal/release/cross"
 	"github.com/nestoca/joy/internal/release/filtering"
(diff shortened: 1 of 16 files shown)


pkg/catalog/catalog.go
```diff
@@ -56,9 +57,23 @@ func Load(dir string, validChartRefs []string) (*Catalog, error) {
 		return nil, fmt.Errorf("matching files with glob expression %s: %w", globExpr, err)
 	}
 
+	// Load .joyignore if it exists
+	ignoreFilePath := filepath.Join(dir, ".joyignore")
```
davidmdm, line 61: You do not need to stat here. It would read much nicer if:
```go
joyIgnore, err := ignore.CompileIgnoreFile(filepath.Join(dir, ".joyignore")
if err != nil && !errors.Is(err, os.ErrNotExists) {
  return nil, fmt.Errror(...)
}
```

davidmdm (approved):

davidmdm (approved):

PR #181, opened 2024-05-28
feat/PL 2736/render many releases

davidmdm:
- **feat(PL-2736): joy release render support all and all-env flags**
- **fix(PL-2736): include both git-ref and diff-ref catalog in known state**

With this PR we will be able to view the diff of the entire catalog's releases rendered via one command:
```
joy release render --all-envs --all --diff-ref REF
```

### DIFF
--- a/internal/release/render/render.go
+++ b/internal/release/render/render.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
-	"io"
 	"os"
 	"regexp"
 	"slices"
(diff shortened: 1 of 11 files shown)


pkg/catalog/catalog.go
```diff
@@ -156,6 +156,23 @@ func (c *Catalog) WithReleaseFilter(filter filtering.Filter) *Catalog {
...
+	c.Releases.Items = []*cross.Release{}
+
+	for _, cross := range releases {
+		if slices.Contains(names, cross.Name) {
+			c.Releases.Items = append(c.Releases.Items, cross)
+		}
+	}
+
+	return c
+}
```
silphid, line 174: Not sure I like that. It's a mutating function disguised into a function that would be returning a filtered copy. I would either rename it to something like `KeepOnlyReleases()` and remove the return value, or more ideally make it deeply clone the catalog.

internal/release/render/render.go
```diff
@@ -360,29 +309,6 @@ func resolveObjectValue(values map[string]any, path []string) (any, error) {
...
-			continue
-		}
-		if strings.HasPrefix(line, "---") || strings.HasPrefix(line, "# Source:") {
-			lines[i] = color.InYellow(line)
-		}
-	}
-
-	n, err := w.dst.Write([]byte(strings.Join(lines, "\n")))
-	return min(n, len(data)), err
-}
```
silphid, line 384: So colouring is officially gone? 😢

silphid (approved):
Pre-approving, but check out my comment about mutating function.

internal/release/render/render.go
davidmdm: Coloring is not gone! It has moved.

pkg/catalog/catalog.go
davidmdm: I did not change the names but I stopped returning the reference, making the API non-chainable and breaking any possible illusion that any cloning is happening.

PR #185, opened 2024-05-30
fix(PL-660): stricter crd decoding

davidmdm:
We should use strict decoding of yaml documents when parsing catalog CRD-Like objects. This way if users add unknown properties such as typos, they can be caught early.

In the current catalog many projects have the typo `repositorySubPaths` instead of the expected `repositorySubpaths`, not `Paths` instead of `paths`.

This means that for some projects the git diff and git log commands do not work as expected.


silphid (approved):
👍

PR #191, opened 2024-06-13
feat(PL-660): add v1alpha1 schemas

davidmdm:


cmd/joy/release.go
```diff
@@ -914,3 +919,55 @@ func NewReleaseLinksCmd() *cobra.Command {
...
+			}
+
+			schema := cuecontext.New().CompileBytes(schemaBytes).LookupPath(cue.MakePath(cue.Def("values")))
+
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), internal.StringifySchema(schema))
+			return err
+		},
+	}
+
+	cmd.Flags().StringVar(&env, "env", "qa", "environment to find release from")
```
silphid, line 971: "qa" seems arbitrary (and nesto-specific). Could either make it a required flag or default to "" and prompt user?

silphid (approved):
LGTM, just one comment!

cmd/joy/release.go
davidmdm: Good catch. I forgot about the nesto-specificness of this for a hot minute. Yes, i will make it required if a release is present

silphid (approved):

PR #193, opened 2024-07-23
feat(PL-2814): Support promoting chart version in joy build promote

alexstojda:


cmd/joy/build.go
```diff
@@ -21,13 +21,15 @@ func NewBuildCmd() *cobra.Command {
...
+	var chartVersion string
+
 	cmd := &cobra.Command{
 		Use:   "promote",
 		Short: "Promote a project to given version",
 		Long: `Promote a project to given version in given environment.
 Typically called at the end of a CI pipeline to promote a new build to default target environment.
 
-Usage: joy build promote <env> <project> <version>`,
+Usage: joy build promote [--chart-version <version>] <env> <project> <version>`,
```
j-martin, line 32: We need this feature because we create a new chart for every new deployment?

If not (or regardless), I would tweak the usage to specify the chart version may not be the same as the version. 

```
[--chart-version <chart-version>]
```

j-martin (approved):

cmd/joy/build.go
alexstojda: Correct, we're publishing a new chart with every release of the service with how I've setup the pipeline right now, similar to how `joy-generator` is also setup.

alexstojda:
```suggestion
Usage: joy build promote [--chart-version <chart-version>] <env> <project> <version>`,
```

PR #196, opened 2024-07-26
feat(PL-2834): remove valuesMapping feature

davidmdm:
## Changes
- Removes config.ValuesMappings

## Why?

mappings are now declarable under chart data. Mappings map values from the release (or static values) into the values object for the chart. This implies that mappings are specific to charts and does not make sense at a global level.

We are removing this feature as it is unwise and confusing to have global configurations that could create conflicts and issues with third party charts down the road.


alexstojda (approved):
Content LGTM, but this should include a `BREAKING CHANGE` mention in the commit for versioning.

silphid (approved):

PR #197, opened 2024-08-23
refactor(PL-2848): Remove all usage of Jac

alexstojda:
- This is a breaking change, but we haven't agreed on when to release v1 of the cli, so its not marked as such.


api/v1alpha1/schemas.cue
```diff
@@ -60,7 +60,7 @@ package v1alpha1
 	metadata!:  #metadata
 	spec: {
 		// Owners is the list of identifiers of owners of the project.
-		// It can be any string that uniquely identifies the owners, such as email addresses or Jac group identifiers.
+		// It can be any string that uniquely identifies the owners, such as email addresses or backstage group names
```
j-martin, line 63: Is the plan to wait for backstage to be available before merging this?

j-martin (approved):

api/v1alpha1/schemas.cue
alexstojda: Nope, the field is now an array of arbitrary strings, there is no actual connection to backstage, its really just a "suggested use".

PR #205, opened 2024-11-12
chore(depot): build with depot

j-martin:
We spend 2m30 out of 4 min caching the post-build. Depot should be much faster than that.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   test-and-build:
-    runs-on: ubuntu-24.04
+    runs-on: depot-ubuntu-24.04
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


alexstojda (approved):

ayahajar:
@jm why not use 24 right away?

j-martin:
Because I copy-pasted it from somewhere else. :P

silphid (approved):

.github/workflows/build-test.yaml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   test-and-build:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-22.04-2
```
ayahajar, line 18: depot-ubuntu-24.04-2

ayahajar (changes requested):

ayahajar (approved):

j-martin:
Closing since it would involve allowing any external PRs to run against depot.

PR #218, opened 2025-04-29
fix(build): use proper token

j-martin:
Note that we may need/want to use a different GH app to publish.


alexstojda (approved):

j-martin:
I have checked the app permission. It should be fine.

PR #224, opened 2025-07-28
fix(PL-3359): support merge tags

davidmdm:


silphid (approved):
LGTM, but PR description should at least mention why! 🙏

PR #259, opened 2026-04-27
fix(PL-5476): handle empty output from gh label list --json

silphid:
# Description & Business motivation

As reported by @adrian.garcia in https://nestoca.atlassian.net/wiki/spaces/DEVOPS/pages/3277291535/Tool+Joy, when calling joy pr promote on a report with absolutely no labels defined gives that error:

```
setting promotion for branch your-branch-name pull request to "staging" environment: adding label staging to branch your-branch-name: unmarshaling label list: unexpected end of JSON input
```

That’s because `gh label list --json ...` returns an empty output when there are no labels defined in the repo, which is invalid JSON.

### DIFF
--- a/internal/github/pull_request_provider.go
+++ b/internal/github/pull_request_provider.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


internal/github/pull_request_provider.go
adriangarcia-nesto, line 1: I'm not familiar with this repository. Should we add/modify some tests accordingly?

silphid: We have integration tests for that command, however I'm not sure it would be worth creating a dedicated test repo (without any labels) just to test that edge case! 😅

davidmdm (approved):

PR #268, opened 2026-05-19
fix(ci): update github workflows

nwaller-nesto:
## What

- Fix script injection: move `github.head_ref` out of inline script interpolation into an env var (`HEAD_REF`)
- Quote the env var usage (`"$HEAD_REF"`) to prevent word splitting
- Pin `nestoca/conventional-changelog-action` to a commit SHA instead of a mutable branch ref (`releases/v4`)

## Why
Two findings from `poutine analyze_local .`:
- **warning** `injection`: `${{ github.head_ref }}` was interpolated directly into a shell script, allowing a malicious branch name to inject arbitrary shell commands
- **note** `github_action_from_unverified_creator_used`: action was pinned to a mutable branch ref; SHA pinning reduces supply chain risk

## How
Fixed with Claude Code

- `poutine analyze_local .` → identified findings → applied fixes → verified clean run for the injection finding

## Test Plan
- [x] Run `poutine analyze_local .` — injection finding is resolved
- [ ] Confirm the `publish` job still works on merge to master (changelog action pinned to same commit it was previously tracking)

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -76,8 +76,9 @@ jobs:
       - name: Dispatch workflow
         env:
           GH_TOKEN: ${{ steps.generate-joy-ci-triggerer-token.outputs.token }}
+          HEAD_REF: ${{ github.head_ref }}
         run: |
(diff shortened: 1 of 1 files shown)


silphid (approved):
OK, interesting, TIL!

PR #277, opened 2026-06-16
feat: add html export

j-martin:
Slippity-slop!

[index.html](https://github.com/user-attachments/files/29014943/index.html)


davidmdm:
This is about the lamest slop I've ever seen.

It couldn't be more creative then literally mapping a table onto a table html element? Where's the minimalist design framework? Where's the spring animations and caroussel?

PR #287, opened 2026-07-20
feat(PL-6895): support custom tags on mapping keys

davidmdm:
Joy uses tags as promotion directives when promoting releases across environments.

These tags have worked generally well for us but have come with one drawback: tags are used to store type information and parsing directives for standard yaml parsers.

When we would add a custom tag to a scalar node, it would lose its type information, and fallback to being interpreted as a string.

Hence:

```yaml
--- release yaml
key: !lock true
--- parsed yaml
key: "true" // not the boolean true.
```

This has always been at odds with out strict schema/cue unification, making sure that developers are writing the correct and desired values for the underlying release's chart.

We've managed to get around this by having joy strip custom tags when unmarshalling releases and by always having joy in the hot-path of any integration. For example the joy-generator would maintain its own local copy of the catalog to load it and recompute all values.

Now that we are moving towards the joy-operator, argocd will be responsible for syncing our catalog resources. This means that joy is no longer responsible for parsing the files, and our tags as promotion directives gets in the way of proper yaml parsing.

To mitigate this, we can move the custom tagging to the key node of a given value instead of on the node itself.

This works because keys should always be strings. (Not always the case in yaml, but always the case with kubernetes APIs). 

This doesn't mitigate the issue completely in theory, as sequence nodes can still fall prey to this issue.
However in practice this has yet to occur.

Regardless, this PR also changes the rendering logic to use the release yaml instead of the custom-parsed release, meaning that we can catch these issues ahead of time.

It also adds a new flag during validation to disalllow custom tags on value nodes within a mapping.

It also ships with a tag migrator. 

---

> [!NOTE]
> **Medium Risk**
> Changes core promotion merge/lock semantics and how releases are decoded for render/validate; the tag migrator overwrites release files on disk if run.
> 
> **Overview**
> Promotion directives (`!lock`, `!local`, etc.) are moving from **mapping values to mapping keys** so release YAML stays valid for standard parsers (e.g. Argo CD) without scalars losing boolean/number types.
> 
> **YAML & promotion behavior** is updated end-to-end: merge propagates tags from keys to values internally then strips duplicate value tags on output; locked/TODO detection and local purging respect tags on keys; `FindNodeKeyPair` and version promotion treat `spec.version` as locked when either the key or value is locked.
> 
> **Validate & render** gain `--no-value-tags` (reject legacy tags on values) and `--raw-yaml` (hydrate/render from the on-disk tree via new `Release.FromTree()` instead of tag-stripped unmarshaling).
> 
> A **`cmd/tag-migrator`** utility rewrites the loaded catalog in place, moving supported custom tags from values onto their keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70070b8e41255625d6139c30652ff9a332f5d433. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added `joy release validate` flags for rejecting tags on mapping values and for validating against raw YAML.
  - Added a tag-migration CLI to move supported YAML tags from mapping values to their keys.
- **Bug Fixes**
  - Improved rendering/validation to hydrate values from the parsed release tree (including when using raw YAML).
  - Fixed YAML merge behavior to properly purge mapping-value custom tags and preserve lock/TODO semantics.
  - Promotion now skips updating `spec.version` when it’s locked.
- **Tests**
  - Expanded validation, rendering, merge, and TODO/lock test coverage for keyed tagged mappings.

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -635,6 +635,7 @@ func NewReleaseRenderCmd() *cobra.Command {
 func NewValidateCommand() *cobra.Command {
 	var env string
 	var noRender bool
+	var noMappingValueTags bool
 
 	cmd := &cobra.Command{
 		Use:   "validate [releases...]",
@@ -672,9 +673,10 @@ func NewValidateCommand() *cobra.Command {
 			}
 
 			return validate.Validate(cmd.Context(), validate.ValidateParams{
-				Releases: releases,
-				NoRender: noRender,
-				Helm:     helm.CLI{IO: internal.IO{Out: cmd.OutOrStdout(), Err: cmd.ErrOrStderr(), In: cmd.InOrStdin()}},
+				Releases:              releases,
+				NoRender:              noRender,
+				NoTagsOnMappingValues: noMappingValueTags,
+				Helm:                  helm.CLI{IO: internal.IO{Out: cmd.OutOrStdout(), Err: cmd.ErrOrStderr(), In: cmd.InOrStdin()}},
 				ChartCache: helm.ChartCache{
 					Refs:            cfg.Charts,
 					DefaultChartRef: cfg.DefaultChartRef,
@@ -687,6 +689,7 @@ func NewValidateCommand() *cobra.Command {
 
 	cmd.Flags().StringVarP(&env, "env", "e", "", "environment to select release from.")
 	cmd.Flags().BoolVarP(&noRender, "no-render", "", false, "skips release rendering validation step")
+	cmd.Flags().BoolVarP(&noMappingValueTags, "no-tags-on-mapping-values", "", false, "disallows tags on mapping values")
 	return cmd
 }
 
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -635,7 +635,7 @@ func NewReleaseRenderCmd() *cobra.Command {
 func NewValidateCommand() *cobra.Command {
 	var env string
 	var noRender bool
-	var noMappingValueTags bool
+	var noValueTags bool
 
 	cmd := &cobra.Command{
 		Use:   "validate [releases...]",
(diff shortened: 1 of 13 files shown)


cmd/joy/release.go
```diff
@@ -687,6 +689,7 @@ func NewValidateCommand() *cobra.Command {
 
 	cmd.Flags().StringVarP(&env, "env", "e", "", "environment to select release from.")
 	cmd.Flags().BoolVarP(&noRender, "no-render", "", false, "skips release rendering validation step")
+	cmd.Flags().BoolVarP(&noMappingValueTags, "no-tags-on-mapping-values", "", false, "disallows tags on mapping values")
```
silphid: Really just a nit, but would have personally prefer something shorter:
```suggestion
	cmd.Flags().BoolVarP(&noMappingValueTags, "no-value-tags", "", false, "disallows tags on mapping values")
```

internal/release/validate/validate.go
```diff
@@ -15,10 +15,11 @@ import (
...
 type ValidateParams struct {
-	Releases   []*v1alpha1.Release
-	Helm       helm.PullRenderer
-	ChartCache helm.ChartCache
-	NoRender   bool
+	Releases              []*v1alpha1.Release
+	Helm                  helm.PullRenderer
+	ChartCache            helm.ChartCache
+	NoRender              bool
+	NoTagsOnMappingValues bool
```
silphid: Similarly here, but I mind much less here within the code than for flags:
```suggestion
	NoValueTags bool
```

internal/yml/merge_test.go
```diff
@@ -743,6 +743,24 @@ func TestYmlMerge(t *testing.T) {
...
 			Expected: "{outer: {}}",
 		},
+		{
+			Name:     "locks on key nodes",
+			Src:      "{ !lock key: value }",
+			Dst:      "{}",
+			Expected: "{!lock key: TODO}",
+		},
+		{
+			Name:     "lcoal on key nodes",
```
silphid:
```suggestion
			Name:     "local on key nodes",
```

silphid (approved):
Good job! 👍 Just a few nits and typos...

PR #293, opened 2026-07-27
feat(PL-5900): create/update/delete preview release

silphid:
As part of [PL-5900](https://nestoca.atlassian.net/browse/PL-5900), this adds `joy release preview create` and `joy release preview delete`, which respectively creates a preview copy of a given release, with given transforms, or deletes that copy. It is intended to be invoked from the `preview-service` workflow/action developed in [that PR](https://github.com/nestoca/actions/pull/1006).

[PL-5900]: https://nestoca.atlassian.net/browse/PL-5900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a `joy preview` CLI command group to create and delete preview copies of releases, including versioning, RFC 6902 patch operations, text replacements, and placeholder substitution.
  * Preview copies are now marked with a dedicated preview label.
* **Bug Fixes**
  * Preview-labeled releases are automatically excluded during promotion.
  * Recreating an existing preview updates only its version.
* **Tests**
  * Added coverage for preview create/update/delete, patch parsing/application and validation, YAML metadata preservation, and promotion skipping.

### DIFF
--- a/internal/patch/patch.go
+++ b/internal/patch/patch.go
@@ -1,103 +1,109 @@
-// Package patch applies a subset of RFC 6902 JSON Patch (add / replace / remove) directly
-// against a gopkg.in/yaml.v3 node tree. Operating on nodes (rather than a JSON round-trip)
-// preserves the catalog's custom YAML tags (e.g. !lock) and node styles.
+// Package patch applies RFC 6902 JSON Patch operations to a YAML release tree.
+//
+// The patch itself is applied by a standard JSON Patch library, so serialization to JSON is
+// required — which drops YAML custom tags (e.g. !lock), comments and key order. Those are
+// restored from the original tree afterwards via yml.CopyMetadata. It also exposes a few
+// node-level setters used for built-in transforms that must run before the JSON round-trip.
 package patch
 
 import (
+	"bytes"
 	"fmt"
-	"slices"
-	"strconv"
-	"strings"
 
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	"gopkg.in/yaml.v3"
+	sigsyaml "sigs.k8s.io/yaml"
+
+	"github.com/nestoca/joy/internal/yml"
 )
 
-// Op is a single RFC 6902 operation.
+// Op is a single RFC 6902 operation, stored as its JSON encoding.
 type Op struct {
-	Op    string    `yaml:"op" json:"op"`
-	Path  string    `yaml:"path" json:"path"`
-	Value yaml.Node `yaml:"value" json:"value"`
+	json []byte
 }
 
-// ParseOp decodes a single RFC 6902 op (YAML or JSON object).
+// ParseOp decodes a single RFC 6902 op (YAML or JSON object) into its JSON form.
 func ParseOp(data []byte) (Op, error) {
-	var op Op
-	if strings.TrimSpace(string(data)) == "" {
-		return op, fmt.Errorf("empty patch op")
+	j, err := sigsyaml.YAMLToJSON(data)
+	if err != nil {
+		return Op{}, fmt.Errorf("decoding patch op: %w", err)
 	}
-	if err := yaml.Unmarshal(data, &op); err != nil {
-		return op, fmt.Errorf("decoding patch op: %w", err)
+	if trimmed := bytes.TrimSpace(j); len(trimmed) == 0 || bytes.Equal(trimmed, []byte("null")) {
+		return Op{}, fmt.Errorf("empty patch op")
 	}
-	// Normalize style so JSON-authored values (e.g. flow maps, double-quoted scalars) render as
-	// clean block YAML in the catalog. The encoder still re-quotes any scalar that needs it.
-	clearStyle(&op.Value)
-	return op, nil
+	return Op{json: j}, nil
 }
 
-func clearStyle(node *yaml.Node) {
-	if node == nil {
-		return
+// Apply applies the ops to doc (a document node) and returns the resulting document. The patch is
+// applied via a JSON Patch library; custom tags, comments and key order are then restored from
+// doc, and styles are cleared so the result serializes as clean block YAML. When there are no
+// ops, doc is returned unchanged.
+func Apply(doc *yaml.Node, ops []Op) (*yaml.Node, error) {
+	if len(ops) == 0 {
+		return doc, nil
 	}
-	node.Style = 0
-	for _, child := range node.Content {
-		clearStyle(child)
+
+	docJSON, err := toJSON(doc)
+	if err != nil {
+		return nil, fmt.Errorf("converting release to json: %w", err)
 	}
+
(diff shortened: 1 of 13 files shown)


davidmdm (approved):

davidmdm (changes requested):
Actually, why are we re-implementing json patch as a yaml node operation?

We should just serialize the release to json and use a battle-tested library implementing the json patch spec.

I believe this is the same one as used by kubernetes:
https://github.com/evanphx/json-patch

silphid:
> Actually, why are we re-implementing json patch as a yaml node operation?
> 
> We should just serialize the release to json and use a battle-tested library implementing the json patch spec.
> 
> I believe this is the same one as used by kubernetes: https://github.com/evanphx/json-patch

However, that would not round-trip well back to yaml! 🤷

davidmdm:
yaml/json, etc. is just a serialization format.

Details and real libs omitted for brevity but it should be straightforward and self-contained, without us having to ship a custom jsonpatch implementation within joy.

```go
func (release *Release) Patch(patches []jsonpatch.Operation) *Release {
  data, _ := json.Marshal(release)
  data = patchTheData(data, patches)
  var result Release
  json.Unmarshal(data, &result)
  return result
}
```

There's also straight yaml to json conversion via `sigs.k8s.io/yaml` that we already include in the project now:
```
	sigsyaml.YAMLToJSON(y []byte)
```

Meaning if we don't want to use the json tags directly, we can still yaml.Marshal, then convert to json in memory, apply the patches, and use yaml to Unmarshal back into a new release since yaml is a superset of json.

alexstojda:
Removing my review request since @davidmdm is already on it :)

internal/patch/patch.go
```diff
@@ -1,103 +1,109 @@
...
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	"gopkg.in/yaml.v3"
+	sigsyaml "sigs.k8s.io/yaml"
+
+	"github.com/nestoca/joy/internal/yml"
 )
 
-// Op is a single RFC 6902 operation.
+// Op is a single RFC 6902 operation, stored as its JSON encoding.
 type Op struct {
```
davidmdm, line 22: This Op type is useful in and of itself, given that the underlying library doesn't really type operations. 
Value would be `any` instead of a `yaml.Node`.

No action required but just a thought.

internal/patch/patch.go
```diff
@@ -1,103 +1,109 @@
...
-	Op    string    `yaml:"op" json:"op"`
-	Path  string    `yaml:"path" json:"path"`
-	Value yaml.Node `yaml:"value" json:"value"`
+	json []byte
 }
 
-// ParseOp decodes a single RFC 6902 op (YAML or JSON object).
+// ParseOp decodes a single RFC 6902 op (YAML or JSON object) into its JSON form.
 func ParseOp(data []byte) (Op, error) {
```
davidmdm, line 30: if the Op was typed we probably wouldn't need a public parse op function.
Plus this function doesn't actually parse an operation, it just converts yaml data to json and makes sure its non-empty.

internal/patch/patch.go
```diff
@@ -1,103 +1,109 @@
...
-		return err
+
+	// The JSON round-trip drops custom tags, comments and key order, and flow-flattens styles.
+	yml.CopyMetadata(&patched, doc)
+	clearStyle(&patched)
+	return &patched, nil
+}
+
+// patchArray concatenates the ops' JSON objects into a single JSON Patch array.
+func patchArray(ops []Op) []byte {
```
davidmdm: This code is odd. 

Why are we modeling patches as structs with a field of json bytes? That then needs to be collected and then surrounded with byte operations to recreate an array?

If we want to continue being typeless, then Ops should be a `[]json.RawMessage` or we can model the types of Op and have a `[]Op` that we marshal to json before use.

internal/patch/patch.go
```diff
@@ -0,0 +1,187 @@
...
+func toJSON(node *yaml.Node) ([]byte, error) {
+	data, err := yaml.Marshal(node)
+	if err != nil {
+		return nil, err
+	}
+	return sigsyaml.YAMLToJSON(data)
+}
+
+// Scalar builds a plain string scalar node.
+func Scalar(value string) *yaml.Node {
```
davidmdm, line 90: Are the functions below used anymore or need to be cleaned up?

internal/yml/copy_metadata.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	dst.HeadComment = src.HeadComment
+	dst.LineComment = src.LineComment
+	dst.FootComment = src.FootComment
+
+	switch dst.Kind {
+	case yaml.DocumentNode, yaml.SequenceNode:
+		for i := 0; i < min(len(dst.Content), len(src.Content)); i++ {
+			CopyMetadata(dst.Content[i], src.Content[i])
+		}
```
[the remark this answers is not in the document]
davidmdm: The bot is probably right here. Maybe for sequences we don't have a great way of preserving metadata. 

Probably omit this for now as a current limitation.

internal/patch/patch.go
silphid: Anything specific to action here? Maybe rename the func?

internal/patch/patch.go
```diff
@@ -0,0 +1,187 @@
...
+	Value any    `json:"value,omitempty"`
+}
+
+// ParseOp decodes a single RFC 6902 op from YAML or JSON.
+func ParseOp(data []byte) (Op, error) {
+	var op Op
+	if err := sigsyaml.Unmarshal(data, &op); err != nil {
+		return Op{}, fmt.Errorf("parsing patch op: %w", err)
+	}
+	if op.Op == "" || op.Path == "" {
```
davidmdm, line 35: The code is overly defensive, and there's a lot I would simply remove. Like I don't see the point of this entire function. But okay.

internal/preview/preview.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		if len(crossRelease.Releases) == 0 || crossRelease.Releases[0] == nil || crossRelease.Releases[0].File == nil {
+			return nil, fmt.Errorf("release %q not found in environment %q", name, env)
+		}
+		return crossRelease.Releases[0], nil
+	}
+	return nil, fmt.Errorf("release %q not found", name)
+}
+
+func applyReplacements(text string, replacements []Replacement) (string, error) {
```
davidmdm, line 175: I am not sure that regex replacements across the entire document is the move. Feels needlessly sensitive to errors. But perhaps this can be revisited.

davidmdm (approved):

PR #295, opened 2026-07-30
Parallelize release render across releases and environments

nwaller-nesto:
## What

`joy release render --all --all-envs` rendered every release/environment combination one at a time. This change parallelizes that render loop and eliminates a redundant CUE schema compilation that was happening on every single render.

1. **Parallelize the render loop.** The nested environment × release loop in `cmd/joy/release.go` now dispatches work through a bounded `errgroup` worker pool (capped at `runtime.NumCPU()`) instead of rendering sequentially.
2. **Cache compiled chart schemas.** `values.cue` was being recompiled from scratch on every single release render, even though many releases share the same chart and therefore the exact same schema. Compilation is now cached per chart, guarded by a per-entry mutex since CUE values from the same context are not safe for concurrent use — only the (already cheap) per-release unification step is serialized per chart, not the expensive compile step.

Also fixes a check-then-pull race introduced by parallelizing: `ChartCache.GetReleaseChartFS` could have concurrent renders race on pulling/untarring the same chart version into the on-disk cache. This is now guarded by a mutex around the check-then-pull sequence.

## Why

Rendering scales with the number of release/environment combinations, and catalogs can have hundreds of these. Sequential rendering means this scales linearly with catalog size; parallelizing lets it scale with available CPU cores instead.

## Performance

Measured against a representative catalog (544 release/environment combinations) on a 12-core machine. Output is byte-identical to the sequential baseline in all cases.

| Stage | Wall time | Speedup |
|---|---|---|
| Baseline (sequential) | 23.45s | 1.0x |
| + Parallelized render loop | 6.56s ± 0.19s | 3.5x |
| + Chart schema caching | 5.19s ± 0.13s | 4.5x |

Profiling showed that after parallelizing, ~29% of total CPU time was spent recompiling identical `values.cue` schemas across releases sharing a chart. Caching that compilation removed most of that cost, shifting the remaining bottleneck to the inherently per-release CUE unification step.

## How

- `golang.org/x/sync` promoted from an indirect to a direct dependency (already present transitively) to bring in `errgroup`.
- Verified with `go test -race` across the touched packages.

### DIFF
--- a/cmd/joy/release.go
+++ b/cmd/joy/release.go
@@ -7,8 +7,10 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"runtime"
 	"slices"
 	"strings"
+	"sync"
 
 	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/cuecontext"
@@ -18,6 +20,7 @@ import (
 	"github.com/davidmdm/x/xerr"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/term"
 
 	"github.com/nestoca/joy/api/v1alpha1"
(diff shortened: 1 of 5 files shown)


davidmdm (commented):
Of the three optimizations here are my thoughts:
1. Errgroup over release rendering: This is worthwhile and the change is minimal
2. Schema caching: seems overly complex. Also unclear to me that cue parsing is the bottleneck versus an exec to `helm template`. I would likely omit this unless it offers measurable and non-negligible improvements.
3. Chart pull serialization now that we are doing concurrency: implemented like a potato. We should not have mutexes in global state like that. Regardless, we've encountered this problem multiple times now: in the joy-generator (deprecated), joy-operator, and in testchart. Given that `helm.Puller` is just an interface, we could probably expose public wrapper that builds in this serialization/concurrency support. Then our other projects could pull this in. (or it could become part of the default implementation of our `helm.CLI` type)

nwaller-nesto:
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closing per discussion with @davidmdm — parallelization is worth keeping but this implementation isn't. We'll rewrite by hand: keep the errgroup-based parallel render loop, drop the CUE schema caching, and redesign the chart-pull concurrency fix as a reusable wrapper (rather than a global mutex) that joy-generator, joy-operator, and testchart can also adopt.

PR #297, opened 2026-08-11
fix: warn when updating an existing preview skips patches/replacements

quentinackermannnexapp:
Create() on an existing preview only bumps spec.version, patches/replaces never get re-applied — intentional (#293), but silent. Spent a chunk of this afternoon debugging why an annotation fix on backoffice#4396 "wasn't taking": the preview file already existed, so every push just re-stamped the version and kept redeploying the old broken annotation underneath. Added a print so the next person doesn't lose the same hour. Delete+recreate is still the only way to actually apply new patches, this just says so out loud instead of staying quiet about it.

go test ./internal/preview/... — added a case for both the warn and no-warn paths.

### DIFF
--- a/internal/preview/preview.go
+++ b/internal/preview/preview.go
@@ -68,6 +68,12 @@ func Create(params CreateParams) error {
 	targetPath := filepath.Join(filepath.Dir(source.File.Path), target+".yaml")
 
 	if _, err := os.Stat(targetPath); err == nil {
+		if len(params.Patches) > 0 || len(params.Replaces) > 0 {
+			fmt.Printf(
+				"%s existing preview %s: only spec.version is updated, patches and replacements are not re-applied. Delete and recreate the preview to pick up changes to those.\n",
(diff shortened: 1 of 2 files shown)


davidmdm:
I wonder if the fix couldn't be to simply apply the patches every time. Instead of opting to only apply the version?

internal/preview/preview.go
```diff
@@ -68,6 +68,12 @@ func Create(params CreateParams) error {
 	targetPath := filepath.Join(filepath.Dir(source.File.Path), target+".yaml")
 
 	if _, err := os.Stat(targetPath); err == nil {
```
davidmdm, line 70: counter proposal. Delete all of this.

quentinackermannnexapp:
Superseded by #299 — David's fix removes the update-only shortcut entirely, so there's nothing left for this warning to guard against. Closing in favor of that.

PR #298, opened 2026-08-11
feat: merge patch for preview transforms

quentinackermannnexapp:
Stacked on #297 (same CreateParams area).

Adding preview releases sometimes need to set a key with a slash in it (e.g. `argocd.nesto.ca/sync.prune`). RFC 6902 JSON Patch addresses keys by JSON Pointer path, so a literal `/` needs `~1` escaping. That escaping got mangled on backoffice#4396 and produced an invalid k8s annotation name apiserver silently rejected.

Preview create now takes a `--merge-patch` flag (RFC 7386) alongside the existing `--patch`/`--replace`. A merge patch is just the target shape in plain YAML, no path syntax, so a slash in a key is a normal map key, no escaping needed. Runs after patches, before replacements. Existing flags unchanged.

go test ./internal/patch/... ./internal/preview/... passes.

### DIFF
--- a/cmd/joy/release_preview.go
+++ b/cmd/joy/release_preview.go
@@ -20,7 +20,8 @@ func NewReleasePreviewCmd() *cobra.Command {
 
 A preview is a copy of a source release named <release><suffix>, pinned to a version and
 labelled ` + "`joy.nesto.ca/preview`" + ` so that "joy build promote" excludes it. Callers layer
-their own transforms via repeatable --patch (RFC 6902 op) and --replace (regex) flags.`,
+their own transforms via repeatable --patch (RFC 6902 op) flags, a single --merge-patch
+(RFC 7386) flag, and repeatable --replace (regex) flags.`,
 	}
 	cmd.AddCommand(newReleasePreviewCreateCmd())
 	cmd.AddCommand(newReleasePreviewDeleteCmd())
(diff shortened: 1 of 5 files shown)


davidmdm (approved):
Approving because I like merge-patch support. However, I am unconvinced about the bug motivating this in the description.

quentinackermannnexapp:
Ok, so in [the original](https://github.com/nestoca/backoffice/pull/4396) pr to test the workflow in backoffice, I used path: /metadata/annotations/'[argocd.nesto.ca/sync.prune](http://argocd.nesto.ca/sync.prune)'
I got an issue with the quotes there because it was literal chars, not YAML syntax
When k8s tried to create annotation key '[argocd.nesto.ca/sync.prune](http://argocd.nesto.ca/sync.prune)' (with quotes), it was an invalid key name

To fix the drops quotes, escapes slash instead: [argocd.nesto.ca](http://argocd.nesto.ca/)~1sync.prune
~1 decodes to / per JSON Pointer spec (so the real annotation becomes: [argocd.nesto.ca/sync.prune](http://argocd.nesto.ca/sync.prune)
This is valid and k8s accepts it. This is just nicer ergonomics so nobody has to hand-escape slashes in keys again - but maybe this was just mb, feel free to close my pr in this case (edited)

PR #300, opened 2026-08-13
fix(PLT-5890): previews auto add preview label and refactorings

davidmdm:
> [!NOTE]
> **Medium Risk**
> Changes core catalog YAML I/O and preview/promote write paths used across CLI workflows; behavior shifts (literal replaces vs regex, encode-on-write) could affect edge-case formatting or custom replace patterns.
> 
> **Overview**
> Refactors how Joy loads, edits, and writes catalog YAML so the **node tree is the source of truth**, and tightens **preview release** creation around the preview label and transforms.
> 
> **`yml.File`** no longer keeps a separate raw `Yaml` buffer or `UpdateYamlFromTree`; bytes are produced on write via **`Yaml()`** / **`MustYaml()`** from `Tree`. Disk writes and loaders (`NewEnvironment` / `NewProject`, promote diff, tag migrator, etc.) are updated to match. Path handling for dotted keys moves to **`yml.SplitIntoPathSegments`** (escaped dots/backslashes), shared with render chart mappings.
> 
> **Preview create** now sets built-ins with **`yml.SetOrAddNodeValue`** (name, **`joy.nesto.ca/preview`** label, `spec.version`), applies caller **RFC 6902** patches through a generic **`patch.Apply`**, then **`yml.CopyMetadata`** (which also clears flow styles) before marshaling. **`--replace`** is **literal** `bytes.ReplaceAll` (plus `__RELEASE__` / `__SUFFIX__`), not regex; **`--patch`** flags unmarshal YAML directly into **`patch.Op`**.
> 
> **`patch`** drops YAML-node helpers (`SetPath`, `ParseOp`, metadata restore inside Apply); promotion/secret import paths stop syncing YAML before write and rely on tree mutation + encode on save.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc3ecad34f6793ed7e2fafdbb4bbf9ba2b706ae2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Preview patch specifications can now be provided directly in YAML format.
  - Patch paths support the `patch` field name.
  - Preview replacements support literal URL prefixes and release/suffix placeholders.
  - YAML files are generated consistently from their latest changes.

- **Bug Fixes**
  - Improved preview generation preserves metadata and produces cleaner, more consistent YAML output.
  - Reduced unexpected formatting differences in patched preview files.
  - Improved handling of YAML paths containing escaped dots and backslashes.


internal/patch/patch.go
```diff
@@ -13,36 +13,22 @@ import (
...
+	Op    string `json:"op" yaml:"op"`
+	Path  string `json:"path" yaml:"patch"`
+	From  string `json:"from,omitempty" yaml:"from,omitempty"`
+	Value any    `json:"value,omitempty" yaml:"value,omitempty"`
 }
 
 // Apply applies the ops to doc (a document node) and returns the resulting document. The patch is
 // applied via a JSON Patch library; custom tags, comments and key order are then restored from
 // doc, and styles are cleared so the result serializes as clean block YAML. When there are no
 // ops, doc is returned unchanged.
```
quentinackermannnexapp: **thought:** this comment still says custom tags/comments/key order get restored and styles cleared "then", inside `Apply`

That logic moved out though, caller does `yml.CopyMetadata` manually now

internal/preview/preview.go
```diff
@@ -67,42 +66,58 @@ func Create(params CreateParams) error {
...
-		return fmt.Errorf("setting preview label: %w", err)
-	}
-	if err := patch.SetPath(node, []string{"spec", "version"}, patch.Scalar(params.Version)); err != nil {
-		return fmt.Errorf("setting spec.version: %w", err)
+	source.Name = target
+	source.Spec.Version = params.Version
+	if source.Labels == nil {
+		source.Labels = map[string]string{}
 	}
+	source.Labels[v1alpha1.PreviewLabel] = "true"
```
quentinackermannnexapp: **question:** is this dead?

everything the function writes out comes from `yml.Clone(source.File.Tree)` a few lines down, not from these typed fields

I don't see anything downstream reading `source.Name`/`Spec.Version`/`Labels` after this point, and `source` is the live pointer from `findSourceRelease` straight out of the catalog

internal/patch/patch.go
```diff
@@ -13,36 +13,22 @@ import (
...
-	var op Op
-	if err := sigsyaml.Unmarshal(data, &op); err != nil {
-		return Op{}, fmt.Errorf("parsing patch op: %w", err)
-	}
-	if op.Op == "" || op.Path == "" {
-		return Op{}, fmt.Errorf(`patch op must have both an "op" and a "path"`)
-	}
-	return op, nil
+	Op    string `json:"op" yaml:"op"`
+	Path  string `json:"path" yaml:"patch"`
```
quentinackermannnexapp: **question:** is the yaml tag on `Path` a typo?

`Path  string \`json:"path" yaml:"patch"\`

json says path, yaml says patch. `--patch` flag parses through `yaml.Unmarshal`, so with this tag `Path` never gets set from user input

I grepped in the repo: `{op: replace, path: /spec/values/ALPHA, value: A2}` parses to `Path:""`
then `patch.Apply` fails with `replace operation value must be object or array`

Am I missing something or should this just be `yaml:"path"`?

quentinackermannnexapp (approved):

quentinackermannnexapp:
**praise**: Great cleanup! 👏 👏 👏

internal/patch/patch.go
davidmdm: I don't read comments so i don't fix them.

I'll fix it.

internal/preview/preview.go
davidmdm: Yes is dead. Good cathc!

internal/patch/patch.go
davidmdm: another great cathc

